PR #9, opened 2015-12-25
added make target coverage

krichter722:
which allows overview of coverage after make check or failure if a certain coverage isn't reached.

Summary of our conversation from https://github.com/krichter722/check-old/pull/1:

@brarcher:
- Adding these files to .gitignore might be better as a separate commit
- It may be better to check if the gcovr tool is installed and at the expected version in the configure script. Further, it may be useful to configure the coverage target as a configure option.
- Is there a way to perform this same check using gcov directly in a few lines, or is gcovr necessary?
Do you happen to know if gcovr is currently being distributed in a package for any linux distro?
- There is a configure options, --enable-gcov, which enables the necessary gcov flags.
Perhaps the coverage target should depend on the check target.

@krichter722: 
> There is a configure options, --enable-gcov, which enables the necessary gcov flags.

I don't understand what it does and it's not documented in `configure --help`.

> Perhaps the coverage target should depend on the check target.

That alone doesn't do it since the test compilation needs to be done with `CFLAGS='-g -O0 --coverage'`.

> Is there a way to perform this same check using gcov directly in a few lines, or is gcovr necessary?

As far as I understand (this is the first time I use `gcov` and `gcovr`) one needs to visualize the results anyway. Furthermore I added a customization of `gcovr` at https://github.com/krichter722/gcovr which allows `gcovr` to return a non-zero code if a certain coverage isn't met. I need to add a check that this customization is installed. Is that possible with `gcov` only?

> Adding these files to .gitignore might be better as a separate commit

Done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@ exported.sym
 stamp-h1
 stamp-vti
 test_vars
+
+# gcovr files
+src/*.gcno
+tests/*.gcda
+tests/*.gcno
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,6 +36,12 @@ pcdata_DATA = check.pc
 
 DISTCLEANFILES = check_stdint.h
 
+# the gcovr binary to use
+GCOVR ?= gcovr
+# the minimal coverage for source files tested with `make check` in the
+# `coverage` target to be accepted (`make coverage` fails otherwise)
+GCOVR_COVERAGE_MIN ?= 100
+
 ACLOCAL_AMFLAGS = -I m4
 
 doc/check_html:
@@ -51,5 +57,10 @@ prereleasecheck: doc/check_html doc/doxygen
 	autoreconf -i && ./configure \
 	    && ulimit -c 0 && \
 	    $(MAKE) distcheck
+coverage:
+	$(MAKE) clean # clean first because change of CFLAGS isn't recognized
+	$(MAKE) CFLAGS='-g -O0 --coverage'
+	$(MAKE) check
+	$(GCOVR) -coverage-min=$(GCOVR_COVERAGE_MIN) . || echo "Make sure you have at least 3.2-11-g7390fa3 of https://github.com/krichter722/gcovr.git installed"
 
 .PHONY: prereleasecheck
--- a/configure.ac
+++ b/configure.ac
@@ -171,6 +171,17 @@ case "${host_os}" in
     ;;
 esac
 
+AC_CHECK_PROG(GCOVR_CHECK,gcovr,yes)
+if test x"$GCOVR_CHECK" != x"yes" ; then
+    AC_MSG_ERROR([Please install gcovr costumization from https://github.com/krichter722/gcovr before building.])
+fi
+if bash -c 'TEMP=$(mktemp -d); touch $TEMP/empty.py; gcovr -coverage-min=0.0 $TEMP'
+then
+  AC_MSG_NOTICE([gcovr costumization installed and working])
+else
+  AC_MSG_ERROR([Please install gcovr customization from https://github.com/krichter722/gcovr before building.])
+fi
+
 AC_CHECK_PROGS(GCOV, gcov, false)
 AC_CHECK_PROGS(LCOV, lcov, false)
 AC_CHECK_PROGS(GENHTML, genhtml, false)


brarcher:
>>There is a configure options, --enable-gcov, which enables the necessary gcov flags.

>I don't understand what it does and it's not documented in configure --help.

It may be that documentation in the output of "./configure --help" is not as helpful as it could be. Curently it says:

```
--enable-gcov           turn on test coverage [default=no]
```

Perhaps I can update this. Would the following be more helpful?

```
--enable-gcov           turns on test coverage with gcov by adding the necessary flags to the build
```

krichter722:
> Perhaps I can update this. Would the following be more helpful?

I think it could go further and explain how to actually reveal the test coverage in percent. That's not `check`'s business, but speaking of helpfulness...

krichter722:
Thanks for your feedback. I corrected the PR according to your proposals. Regarding the SVN issue I simply removed the extranous lines (I created the first check repository with `git-svn` which created commits which have nothing to do with those of this repository and had to diff and patch manually).

krichter722:
Regarding pushing changes to `gcovr` upstream: I can only open a pull request and hope for the `gcovr` devs to react to it in any way which hasn't happened yet. For instance the usage of my extension is messy, but it works and is imo the only way because I didn't find any tool which can fail if a coverage `x` isn't met in a C project.

brarcher:
(Sorry for updating your comment, I though I was replying to it instead...).

> > Perhaps I can update this. Would the following be more helpful?

> I think it could go further and explain how to actually reveal the test coverage in percent. That's not `check`'s business, but speaking of helpfulness...

Really, enabling gcov and analyzing unit test coverage is something useful for a developer/contributor of Check, not to a user. I believe that if one is interested in compiling Check and determining its unit test coverage, it is acceptable for one to lookup how to best use gcov.

There is a Jenkins job which runs on each pull request and analyzes the unit test coverage of Check:

https://check.ci.cloudbees.com/job/check-gcov-github/

There is no automatic analysis, and instead one would need to look at the output. It one were to submit a pull request which added new code, I would use the output from that job to determine if any additional unit test coverage were necessary.

Adding the make target for checking coverage is not a bad step. Note though that there are certain parts of the code which are acceptable to not have unit test coverage, and 100% coverage is not obtainable always. For example,:
```
switch(someVar)
{
   case 1: doSomething(); break;
   case 2: doSomethingElse(); break;
   default: reportError(); break;
}
```
It may be that the default case is there, but is never covered in a unit test as it "should never happen". Sure, there are techniques one could use to remove such examples and get close to 100%, but at least as things stand Check does not achieve 100%.

brarcher:
> but it works and is imo the only way because I didn't find any tool which can fail if a coverage x isn't met in a C project.

I am sorry, but unless there is a project which exists that has the functionality of determining the current % of test coverage and can be installed from a package manager (e.g. not from source from github) it does not make much sense to add a dependency to it in Check.

Given how the jenkins job is using gcov, the following quick and dirty shell script would convert the gcov data to a percentage:
```
for file in `ls *.summary.txt`; do cat $file; done | grep "Lines executed" | cut -d ":" -f 2 | tr -d "%" | awk '{checked+=$1*$3/100; total+=$3} END {print checked/total*100}'
```

The current output for Check is ~89%.

krichter722:
I understand that `check` needs to rely on stable dependencies and that it makes sense to wait gcovr to reply to the -coverage-min feature proposal.

> Adding the make target for checking coverage is not a bad step. Note though that there are certain parts of the code which are acceptable to not have unit test coverage, and 100% coverage is not obtainable always.

I see. If would be a quite simple discussion about the percentage, then.

I created this target because I wanted to investigate the `No messaging set up` issue (will add a link if requested), so it's just a by-product and it's written nowhere that this has to be integrated. So, I'd like to leave this open or close it with the option of reopening it.

Instead the above explanation about `gcov` and jenkins should go into `HACKING`.

krichter722:
> for file in `ls *.summary.txt`; do cat $file; done | grep "Lines executed" | cut -d ":" -f 2 | tr -d "%" | awk '{checked+=$1*$3/100; total+=$3} END {print checked/total*100}'

Or just run `gcovr` (is installable in Ubuntu 15.10 with `apt-get install gcovr`). Both could be mentioned in `HACKING`.

> The current output for Check is ~89%.

That's what `gcovr` reveals as well.

brarcher:
> I created this target because I wanted to investigate the No messaging set up issue

Are you referring to this bug:

https://sourceforge.net/p/check/bugs/111/

I'm not sure what might be going on there, but if you are able to create a minimal example, it would help in tracking down the cause.

> Instead the above explanation about gcov and jenkins should go into HACKING.

I'll do that, thanks for the suggestion. Do you think that this is sufficient?

https://github.com/libcheck/check/pull/11

> I'd like to leave this open or close it with the option of reopening it.

That is fine. If it is not going to be actively pursued, I would prefer the close with the option to reopen later.

brarcher:
When you are ready, this can be reopened or a new pull request created.

PR #16, opened 2016-01-08
web: add Arch Linux installation information

ExcaliburZero:
Add information to the installation webpage on how to install the program in Arch Linux using Pacman.

Also fix a small grammar error in the Fedora installation section.

When I was editing the file in Vim I noticed that it seemed to use "^M" to end lines. However, I am unfamiliar with this setup, and was unable to replicate this with some of the lines I modified. Please let me know how I can use this setup, and I will alter this pull request to use it.


brarcher:
Jenkins: ok to test

brarcher:
The ^M in the files are the \r character. I did not realize that my editor was setup to end lines with \r\n. Either way, I am not concerned if your patch ends lines with \n only.

Thanks for the patch!

PR #24, opened 2016-03-21
Fix a type in configure.ac and conform to POSIX strsignal

blueness:
We fix to items here, as per the commit messages:

1)  configure.ac: fix typo.  AC_CHECK_DECLS requires commas between declaration.  The missing comma leads to HAVE_DECL_GETLINE_GETTIMEOFDAY which is non-sensical.

2) lib/strsignal.c: strsignal() should be not be declared const.  POSIX requires strsignal() to return a pointer to a char, not a pointer to a const char. (see man 3 strsignal)  On uClibc, and possibly other libcs, this causes problems with the correct declaration in string.h.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,6 +16,7 @@ Committers:
     
 Contributors:
     Cesar Ballardini      (signals)
+    Anthony G. Basile     (fix configure.ac and strsignal())
     Friedrich Beckmann    (mingw and msvc port #1)
     Frank Bergmann        (WIN32 tmpfile workaround)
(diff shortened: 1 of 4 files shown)


brarcher:
jenkins: ok to test

brarcher:
Thanks for the patches!

If you do not have an entry in the [AUTHORS](https://github.com/libcheck/check/blob/master/AUTHORS) file, can you also add a change to that file and add a line for yourself?

blueness:
okay i've added myself to the AUTHORS file as a contributor

PR #25, opened 2016-03-22
In cmake build, use mkstemp when present

jd-boyd:
This fixes this error, seen when building with cmake:

    check/src/check_msg.c:247: warning: the use of 'tempnam' is dangerous, better use `mkstemp'

`CMakeLists.txt` checks for `mkstemp` being present and sets a cmake
variable `HAVE_MKSTEMP`.  The `check_msg.c` file will use `mkstemp` if
`HAVE_MKSTEMP` is true, but the cmake variable wasn't causing `HAVE_MKSTEMP`
to be defined for the C compiler.

Also, trim some trailing white space.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,7 +172,7 @@ check_symbol_exists(SSIZE_MAX "limits.h"   HAVE_SSIZE_MAX)
 ###############################################################################
 # Check struct members
 
-# Check for  tv_sec in struct timeval 
+# Check for  tv_sec in struct timeval
 if(NOT HAVE_SYS_TIME_H)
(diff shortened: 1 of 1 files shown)


brarcher:
Jenkins: ok to test

brarcher:
> Also, trim some trailing white space.

Can the whitespace changes be put into its own commit?

Also, please rebase the request to master master? One other change went in ahead of this one.

jd-boyd:
> Can the whitespace changes be put into its own commit?
> 
> Also, please rebase the request to master master? One other change went in ahead of this one.

Both are done.

Issue #26, opened 2016-03-22
CMake build fails on OS X


kainjow:
Using Xcode 7.3, check fails to build:

    cmake ../check && make

First error is:

    lib/timer_settime.c:23:5: error: conflicting types for 'timer_settime'
    int timer_settime(timer_t timerid CK_ATTRIBUTE_UNUSED,
         ^
    lib/libcompat.h:196:16: note: previous declaration is here
    CK_DLL_EXP int timer_settime(timer_t timerid, int flags,

OS X doesn't have `itimespec` but the CMake file doesn't properly detect this. If I disable the ifs in CMakeLists.txt on lines 167 and 168 for `if(NOT HAVE_SYS_TIME_H)` and `if(MSVC)` then it compiles. Not sure if this is the right fix though.

Also, the Travis tests don't test CMake, might be good to add.

brarcher:
Currently [Check only supports CMake for Windows](http://libcheck.github.io/check/web/install.html); all other platforms are encouraged to use the provided configure script.

If one were interested in providing changes which get CMake working on OSX, we would be happy to accept them and add something to the travis-ci config for building with CMake on OSX. If you have that interest, I'll leave this issue open.

brarcher:
Looking at your project, I see that at [test time](https://github.com/civetweb/civetweb/blob/641a742128de6330db4ffeddc6c7be96206f5dfd/test/CMakeLists.txt) Check is being downloaded, compiled with CMake, and unit tests run. Probably this worked for Windows and GNU/Linux, but perhaps not other platforms such as OSX.

Originally I added CMake support to Check so that Windows could be supported. CMake also happens to work for GNU/Linux, and there is one job on our [Jenkins instance](https://check.ci.cloudbees.com/job/check-cmake-github/) which verifies that it continues to work. To date no work has gone into support CMake on other platforms, as the autotools build system works just fine. Probably for now Check will not officially support CMake on any platform besides Windows. If it happens to work, though, I'm not opposed to setting up something on our Jenkins instance or on Travis-CI to keep it working. 

If you were able to provide some fixes to Check to enable CMake support for OSX, I can review the changes and add them to Check. If the fixes also happen to apply cleanly to Check 0.10.0, then you may have a way to use the latest Check release with your patches to get your OSX build working now before Check has its next release.

[...]

PR #27, opened 2016-03-23
Change travis config to test both cmake and autoconf

jd-boyd:
Certainly #25 should go in first, but I'm starting this pull request so that I can see the result of the travis run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,5 @@ script:
   - if [ $USE_CMAKE == "NO" ] ; then pushd ../ ; autoreconf -i ; popd; fi
   - if [ $USE_CMAKE == "NO" ] ; then ../configure ; fi
   - make
-  - make check
+  - if [ $USE_CMAKE == "YES" ] ; then CTEST_OUTPUT_ON_FAILURE=1 make test ; fi
+  - if [ $USE_CMAKE == "NO" ] ; then make check ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,21 @@ os:
 compiler:
   - gcc
   - clang
-  
+
+env:
+  - USE_CMAKE=YES
+  - USE_CMAKE=NO
+
 addons:
   apt:
     packages:
       - texinfo
 
 script:
-  - autoreconf -i
-  - ./configure
+  - mkdir build
+  - cd build
+  - if [ $USE_CMAKE == "YES" ] ; then cmake ../ ; fi
+  - if [ $USE_CMAKE == "NO" ] ; then pushd ../ ; autoreconf -i ; popd; fi
+  - if [ $USE_CMAKE == "NO" ] ; then ../configure ; fi
   - make
   - make check
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,12 +13,13 @@ Committers:
     Zdenek Crha           (new Check API docs, fixes, and tests)     |  zdenekc
     Mateusz Loskot        (msvc port #2)                             |  mloskot
     Jose E. Marchesi      (selective testing support)                |  jemarch
-    
+
 Contributors:
     Cesar Ballardini      (signals)
     Anthony G. Basile     (fix configure.ac and strsignal())
     Friedrich Beckmann    (mingw and msvc port #1)
     Frank Bergmann        (WIN32 tmpfile workaround)
+    Joshua Boyd           (travis testing with both build systems, cmake fixes)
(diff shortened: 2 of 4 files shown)


jd-boyd:
OK, this works for cmake and autotools on linux, but the cmake builds on OSX fail.  I see that #26 was already opened for this.  I guess working on that may be where I go next.

jd-boyd:
If #29 is accepted, and this is rebased on top, then it should have completely passing tests.

brarcher:
If you do not have an entry in the AUTHORS file yet, can you include a change to this pull request adding yourself?

.travis.yml
```diff
@@ -7,14 +7,19 @@ os:
...
   apt:
     packages:
       - texinfo
 
 script:
-  - autoreconf -i
-  - ./configure
+  - mkdir build
+  - cd build
+  - ../scripts/run_config.sh
```
brarcher: A small thing, but as the number of lines necessary to build is small, could it be included directly in the .travis.yml file instead of as a separate script?

jd-boyd: I can do that.  

At:
https://groups.google.com/forum/#!topic/travis-ci/uaAP9zEdiCg
one of the travisci founders says:
> The recommended way to do this is to pull out the commands into a shell script and have that be run from your .travis.yml instead.

However, it wasn't quite as ugly to inline as I expected, lets see if it passes.

I'm still planning to rebase/squash after #29 is merged.  If it isn't merged, then more changes will be required to omit cmake testing on osx.

jd-boyd:
> If you do not have an entry in the AUTHORS file yet, can you include a change to this pull request adding yourself?

Will do.

AUTHORS
```diff
@@ -13,12 +13,13 @@ Committers:
...
     Mateusz Loskot        (msvc port #2)                             |  mloskot
     Jose E. Marchesi      (selective testing support)                |  jemarch
-    
+
 Contributors:
     Cesar Ballardini      (signals)
     Anthony G. Basile     (fix configure.ac and strsignal())
     Friedrich Beckmann    (mingw and msvc port #1)
     Frank Bergmann        (WIN32 tmpfile workaround)
+    Joshua Boyd		  (travis testing with both build systems, cmake fixes)
```
brarcher: It is a real nit-pick, but could the () be lined up with the others?

brarcher:
The CMake change for OSX has been merged in. Feel free to update this request.

brarcher:
Jenkins: ok to test

brarcher:
Looking at the build results on Travis, CMake now does work for OSX. The configuration for Travis, however, does not run the unit tests when using CMake. It turns out that for CMake the test target is currently '[test](https://cmake.org/Wiki/CMake/Testing_With_CTest)' instead of 'check'. I think that this is a convention of CMake.

If you could, would you mind in your change running 'make check' for autotools and 'make test' for CMake?

jd-boyd:
> If you could, would you mind in your change running 'make check' for autotools and 'make test' for CMake?

I've pushed a commit with that change and alas it looks like there is more work to do to get CMake full up to speed.

brarcher:
The AppVeyor failures are my fault, as I am working on getting it working in another branch. They can be ignored.

The Travis-CI failures for CMake show that CMake does not let unit tests run on Linux or OSX. However, [Jenkins](https://check.ci.cloudbees.com/job/check-cmake-github/33/console) did show CMake was successful on Linux. The difference is that the Travis-CI build attempted to build in another directory instead of directly in the source tree. Further evidence is the following from the Travis-CI build, showing that the unit testing script cannot find a binary which was built:
```
test_output.sh: 1: test_output.sh: ./ex_output: not found
```
The autotools setup does allow building out of tree, but it appears that CMake does not quite yet. If the CMake changes did not build in a separate build directory, they would probably be successful. If that is sufficient for now, you could make that adjustment a separate defect can be created for letting CMake build out of tree, if it is important.

jd-boyd:
> The autotools setup does allow building out of tree, but it appears that CMake does not quite yet.

CMake strongly recommends building out of tree. I think the problem is in how CMake is invoking the test shell scripts and building test_vars, but I haven't completely gotten it worked out yet.

Also, I noticed that the `check_check_export` test was never added to CMake, so one more todo item.

jd-boyd:
The tests (minus the still not added `check_check_export` test) pass locally, so lets see what Travis makes of them.

brarcher:
Nice! Thanks for helping dig into this.

PR #29, opened 2016-03-23
Detect missing itimerspec on OSX.

jd-boyd:
Set define to compiler accordingly.

This fixes cmake on osx support.


brarcher:
Jenkins: ok to test

Issue #34, opened 2016-03-31
configure in doc/example fails with "cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."" when autoconf is used


krichter722:
`cmake . && make && make test` works fine. `README` suggest that both work fine. Output:

```
$ autoreconf -i -v -f
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force 
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
libtoolize: copying file 'build-aux/ltmain.sh'
libtoolize: Consider adding 'AC_CONFIG_MACRO_DIRS([m4])' to configure.ac,
libtoolize: and rerunning libtoolize and aclocal.
libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
autoreconf: running: /usr/bin/autoconf --force
autoreconf: running: /usr/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:6: installing 'build-aux/compile'
configure.ac:16: installing 'build-aux/missing'
src/Makefile.am: installing 'build-aux/depcomp'
autoreconf: Leaving directory `.'
$ ./configure 
configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."
```

experienced with 0.10.0-69-g82cdb29

brarcher:
My guess is that something necessary for the configure script to run is missing. A quick [Google search](https://acassis.wordpress.com/2012/05/13/configure-error-cannot-find-install-sh-install-sh-or-shtool-in/) points out that you may not have libtool, shtool, and/or autogen installed. If this is the Ubuntu system mentioned in another issue, could you try installing these and see if it resolves the problem?
```
sudo apt-get install libtool
sudo apt-get install shtool
sudo apt-get install autogen
```

As a side question, does this occur using Check 0.10.0 with the provided configure script? How about generating one using autoreconf?

krichter722:
Sorry, I'm an idiot for not mentioning that this happens in `doc/example`, but some of your comment still applies.

> My guess is that something necessary for the configure script to run is missing. A quick Google search points out that you may not have libtool, shtool, and/or autogen installed. If this is the Ubuntu system mentioned in another issue, could you try installing these and see if it resolves the problem?

[...]

PR #36, opened 2016-04-19
Remove AC_CONFIG_AUX_DIR

Earnestly:
* autoconf 2.69
* automake 1.15

This is half an issue report and half a pull request.

The reason for this is that while autoreconf -fi works for most of the
needed scripts it doesn't copy ltmain.sh into build-aux which results
the following ./configure failure:

    configure.ac:137: error: required file 'build-aux/ltmain.sh' not found

This patch, which is a similar one I've provided to cloog[0] which
allows ./configure to at least complete.

I'm not sure what you think since obviously the directory was used for
some reason, likely to reduce top-level directory clutter, but this at
least works for me.

This will also likely solve https://github.com/libcheck/check/issues/34

0 https://groups.google.com/forum/?hl=en#!topic/cloog-development/28_M3Qk87-Q

Signed-off-by: Earnestly <zibeon@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,9 +19,6 @@ AC_CONFIG_MACRO_DIR([m4])
 # place where portability library functions are kept
 AC_CONFIG_LIBOBJ_DIR([lib])
 
-# place to put some extra build scripts installed
-AC_CONFIG_AUX_DIR([build-aux])
-
 # really severe build strictness
 AM_INIT_AUTOMAKE([-Wall gnits 1.11.2])
 # Change to using into-in-builddir in the future:


brarcher:
Jenkins: ok to test

brarcher:
I like issues that come with pull requests (:

Looking up information on why ltmain.sh may be missing, the [automake documentation](https://www.gnu.org/software/automake/manual/html_node/Error-required-file-ltmain_002esh-not-found.html) mentions that the reason may be a change in more recent versions. This is conjecture, but perhaps part of this change which moves the responsibility of invoking libtoolize to autoreconf results in libtoolize not knowing to put ltmain.sh into the config aux dir. That may be a bug in autoreconf (not sure), but simply not using the config aux dir as a workaround seems reasonable.

I'll let the Jenkins build complete. If successful, I'll accept the change.

Earnestly:
Yeah I think your analysis is correct.  GNU seems to be moving away from `./autogen.sh` and `./bootstrap` scripts in favour of using `autoreconf` and this might just be teething issues.

PR #40, opened 2016-04-24
Make variable arguments contain at least one item

xsoameix:
Because adding '##' comma in front of `__VA_ARGS__` is gcc extension,
removing this requirement makes library more close to c89.

This commit is based on latest branch `master`.

### DIFF
--- a/src/check.h.in
+++ b/src/check.h.in
@@ -390,14 +390,14 @@ static void __testname (int _i CK_ATTRIBUTE_UNUSED)\
  *
  * This call is deprecated.
  *
- * NOTE: The space before the comma sign before ## is essential to be compatible
- * with gcc 2.95.3 and earlier.
+ * NOTE: The helper makes the argument list contain at least one item.
  * FIXME: these macros may conflict with C89 if expr is
  * FIXME:   strcmp (str1, str2) due to excessive string length.
  */
-#define fail_if(expr, ...)\
+#define fail_if(...) _fail_if_helper(__VA_ARGS__, NULL)
+#define _fail_if_helper(expr, ...)\
   (expr) ? \
-     _ck_assert_failed(__FILE__, __LINE__, "Failure '"#expr"' occurred" , ## __VA_ARGS__, NULL) \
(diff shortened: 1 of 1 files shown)


brarcher:
Here is one example test which is failing, but should pass, on [MSVC](https://ci.appveyor.com/project/brarcher46583/check/build/1.0.74/job/bprc25ndg13byabt):
```
START_TEST(test_fail_if_pass)
{
  fail_if(1 == 2, "This test should pass");
  fail_if(0, "This test should pass");
}
END_TEST
```

This test should pass both checks, because neither of the conditions are true. However, from the unit testing logs, we see the following line:
```
C:\projects\check\tests\check_check_sub.c:62:F:Simple Tests:test_fail_if_pass:0: Failure '1 == 2, "This test should pass"' occurred
```

I cannot say why this is the case. There are a number of other tests which are also failing. The logs seem to indicate that the majority of the tests are failing.  This can be seen by the lines which start with:
```
test_check_all_msgs:
```
They check that the message emitted from Check is as expected. They should end in Passed, such as:
```
C:\projects\check\tests\check_check_master.c:456:P:Core Tests:test_check_all_msgs:0: Passed
```
However, many end in failure, for example:
```
C:\projects\check\tests\check_check_master.c:475:F:Core Tests:test_check_all_msgs:8: Expected 5 != 6, got Failure 'x != y, "%d != %d", x, y' occurred
```
One may need to determine what the Microsoft compiler is attempting to do and why it is different from what Clang and GCC is attempting.

brarcher:
Are you interested in continuing to pursue this change, @xsoameix?

xsoameix:
Yes, I will make myself available for this pull request.

njh:
I did find this interesting StackOverflow answer:
http://stackoverflow.com/questions/32047685/variadic-macro-without-arguments

Does that help?

brarcher:
As there has been no activity in this merge requests for many months, I'll close it for now. If you are interested in pursuing this change further, please re-open this merge request. Thanks.

Issue #41, opened 2016-05-16
dangerous use of strings as format strings


UlrichEckhardt:
It seems that some strings are carelessly passed to the `printf()` family of functions as format string. The following code should demonstrate the issue:
```
START_TEST (test_somefunction)
{
    // Note: This always fails!
    ck_assert_int_eq(somefunction("%n"), 1);
}
END_TEST
```
The problem shows here because `%n` requires a pointer where it stores the number of written characters. Since none is provided, this reliably causes a segfault:
```
uli@zox ~/devel> ./a.out 
Running suite(s): Test Suite
0%: Checks: 1, Failures: 0, Errors: 1
libcheck-bug.c:12:E:Test Case:test_somefunction:0: (after this point) Received signal 11 (Segmentation fault)
```
In other cases, percent sequences only cause garbled output or other unexpected things. As a fix, either use `printf("%s", dangerous_string)` or `puts(dangerous_string)`.

Full example code:
```
#include <check.h>
#include <stdlib.h>

int somefunction(char const* c)
{
    return 0;
}

START_TEST (test_somefunction)
{
    // Note: This always fails!
    ck_assert_int_eq(somefunction("%n"), 1);
}
END_TEST

int main()
{
    Suite *s = suite_create("Test Suite");
    TCase *tc_core = tcase_create("Test Case");

    tcase_add_test(tc_core, test_somefunction);
    suite_add_tcase(s, tc_core);

    int number_failed;
    SRunner *sr = srunner_create(s);
    srunner_run_all(sr, CK_NORMAL);

    number_failed = srunner_ntests_failed(sr);
    srunner_free(sr);
    return (number_failed == 0) ? EXIT_SUCCESS : EXIT_FAILURE;
}
```

This was initially found on Elementary OS and reproduced later with Debian's version 0.9.10-6.1. BTW: You can add PulseAudio to the list of projects using this library.

brarcher:
I believe that this issue is already resolved. Attempting the example program on master leads to the following output;
```
$ tests/test_safe_printf 
Running suite(s): Test Suite
0%: Checks: 1, Failures: 1, Errors: 0
test_safe_printf.c:12:F:Test Case:test_somefunction:0: Assertion 'somefunction("%n") == 1' failed: somefunction("%n") == 0, 1 == 1
```
Likely [this](https://github.com/libcheck/check/commit/d9c22354c9b7af9191f10a465ed6fa5b77f1de67) commit resolves the issue.

[...]

PR #43, opened 2016-05-26
Add test case where %n is passed into a ck_assert call

brarcher:
In the past it was possible to cause issues if arguments
were passed into ck_assert* macros which included %n.
Exist unit tests check for output format options, such as
%d and %f. No test existed for %n, which writes a value to
memory if used.

Ulrich Eckhardt observed that passing something with a %n in it
would result in issues running Check 0.9.10. Although the
issue is already resolved, this commit adds the test case mentioned in
Ulrich's GitHub issue: https://github.com/libcheck/check/issues/41


brarcher:
Jenkins: ok to test

brarcher:
Jenkins: ok to test

brarcher:
Jenkins: test this please

brarcher:
Jenkins: test this please

PR #45, opened 2016-06-06
Adding an EditorConfig file

RyDroid:


.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+root = true
+
+# LF end-of-line, insert an empty new line and UTF-8
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+# Tab indentation
+[makefile,Makefile]
```
brarcher: Is this to say that only Makefiles should use tabs, whereas everything else is spaces?

.editorconfig
```diff
@@ -0,0 +1,19 @@
+# Copying and distribution of this file, with or without modification,
```
brarcher: Can this use the same license as the rest of the project, namely LGPLv2?

RyDroid:
- The comment was wrong. The rule about tab indentation only applies to makefile(s).
- I think it is using a bazooka to license this very small file without dependency under LGPLv2+. However, I removed the small lax license notice, but the same file can be found in some of my projects under the small non copyleft permissive license,

brarcher:
Thanks for clearing up my questions.

If you would, can you rebase this request against master and push again? I have the project configured to only allow merges which are up-to-date with master.

RyDroid:
Done and thank you for this free/libre framework.

brarcher:
This is a formality to finish kicking off the remaining tests:

Jenkins: ok to test

brarcher:
> thank you for this free/libre framework.

And thank you for your contribution!

PR #46, opened 2016-06-06
Improving README

RyDroid:


brarcher:
This change should have no effect on building, however for formality it will need to go through all the normal checks. The following kicks off the tests hosted on our Jenkins instance:

Jenkins: ok to test

PR #62, opened 2016-09-20
Make TAP test report compatible with standard

eduardoabinader:
Instead of printing 1...N at the end, we print at the beginning of the test, in order to comply to TAP specification.

### DIFF
--- a/src/check_impl.h
+++ b/src/check_impl.h
@@ -127,6 +127,7 @@ struct SRunner
     enum fork_status fstat;     /* controls if suites are forked or not
                                    NOTE: Don't use this value directly,
                                    instead use srunner_fork_status */
+    int num_current_run_tests;
 };
 
 
--- a/src/check_log.c
+++ b/src/check_log.c
@@ -351,12 +351,11 @@ void tap_lfun(SRunner * sr CK_ATTRIBUTE_UNUSED, FILE * file,
             num_tests_run = 0;
             break;
         case CLENDLOG_SR:
-            /* Output the test plan as the last line */
-            fprintf(file, "1..%d\n", num_tests_run);
-            fflush(file);
             break;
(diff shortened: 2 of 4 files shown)


eduardoabinader:
Stuck?

brarcher:
The Jenkins server is setup to start testing when it sees the following comment:

Jenkins: ok to test

The last run of tests should now go through.

brarcher:
To ask about the nature of this change, the [TAP specification](https://testanything.org/tap-specification.html) says:

> The plan tells how many tests will be run, or how many tests have run. It’s a check that the test file hasn’t stopped prematurely. It must appear once, whether at the beginning or end of the output.

My understanding is that having the plan as the last item is still valid according to the specification. Do you agree, or is there another part of the specification that disagrees with this?

brarcher:
One of the test scripts has reported a failure. [The following attempt](https://check.ci.cloudbees.com/job/check-linux-github/87/COMPILER=gcc,FORK=no-fork,HOST=x86_64-redhat-linux,SNPRINTF=system-snprintf,SUBUNIT=no-subunit,TIMER=system-timer/console) shows the following:

```
10:15:50 Problem with ex_tap_output
10:15:50 Expected:
10:15:50 ok 1 - ex_output.c:Core:test_pass: Passed
10:15:50 not ok 2 - ex_output.c:Core:test_fail: Failure
10:15:50 
10:15:50 Got:
10:15:50 1..8
10:15:50 ok 1 - ex_output.c:Core:test_pass: Passed
10:15:50 not ok 2 - ex_output.c:Core:test_fail: Failure
10:15:50 FAIL: test_tap_output.sh
```

eduardoabinader:
Oh yeah, I do agree. That's right.

PR #71, opened 2016-11-06
Allow tests to be checked with Valgrind via ctest

9fcc:
If some internal tests have segmentation faults those could be located with "ctest -D ExperimentalMemCheck".

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -40,6 +40,8 @@ endif(WIN32)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/test_vars.in
   ${CMAKE_CURRENT_BINARY_DIR}/test_vars)
 
+include(CTest)
+
 set(CHECK_CHECK_SOURCES
   check_check_exit.c
   check_check_fixture.c


brarcher:
Currently there is a [test](https://github.com/libcheck/check/blob/master/tests/test_mem_leaks.sh) which runs along with pull-requests tests that executes a specific unit test program under Valgrind to ensure there are no memory leaks. There are some unit tests in Check which do intentionally leak memory, and these are not included in the Valgrind test. The overall goal is to show that the Check framework does not itself introduce memory leaks.

Have you observed locally, perhaps while modifying Check, that sometimes unit tests produce segfaults or memory leaks? Is your change to provide a mechanism for more easily running Valgrind against a unit test program?

I did attempt this change locally but I was unable to get it to do anything:
```
$ ctest -D ExperimentalMemCheck
   Site: Brandens-Computer.local
   Build name: Darwin-
Memory check project /Users/brarcher/workspace/check.git
Memory checker (MemoryCheckCommand) not set, or cannot find the specified program.
Errors while running CTest
```
Does one need something else installed to get this to work?

9fcc:
You need to install valgrind. On debian-based systems:
apt-get install valgrind
cmake finds it automatically.
It can be used to debug your own changes locally.

brarcher:
Jenkins: ok to test

PR #73, opened 2016-11-07
Use ctest on VC builds to run unit tests

brarcher:
msbuild was running the unit tests, but if a failure was hit
the output of the unit tests was not printed. As a result,
there was no indication as to what actually failed in a given
unit test program. Using ctest (part of CMake) should
output the unit test program's output for better failure
diagnosis.


brarcher:
Jenkins: test this please

PR #87, opened 2016-12-13
Add pointer against NULL comparison macros

9fcc:
To fulfil the API set, missing pointer macros.

### DIFF
--- a/doc/check.texi
+++ b/doc/check.texi
@@ -1223,6 +1223,17 @@ used for comparison is different for each function and is indicated by
 the last two letters of the function name.  The abbreviations @code{eq} and
 @code{ne} correspond to @code{==} and @code{!=} respectively.
 
+@item ck_assert_ptr_null
+@itemx ck_assert_ptr_nonnull
+
+Compares a pointers against null and displays predefined message with
+condition and value of the input parameter on failure.
+@code{ck_assert_ptr_null} checks that pointer is equal to NULL and
+@code{ck_assert_ptr_nonnull} checks that pointer is not equal to NULL.
+@code{ck_assert_ptr_nonnull} is highly recommended to use in situations
+when a function call can return NULL as error indication (like functions
+that use malloc, calloc, strdup, mmap, etc).
+
 @item ck_assert_mem_eq
 @itemx ck_assert_mem_ne
 @itemx ck_assert_mem_lt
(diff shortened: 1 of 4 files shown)


brarcher:
Jenkins: ok to test

doc/check.texi
```diff
@@ -1217,6 +1217,17 @@ would pass.
 @item ck_assert_ptr_eq
 @itemx ck_assert_ptr_ne
 
+Compares a pointers against null and displays predefined message with
+condition and value of the input parameter on failure.
+@code{@item ck_assert_ptr_null} checks that pointer is equal to NULL and
```
brarcher: I do not think you need the @item inside the @code{}. This may be what the build is complaining about when checking the documentation.

tests/check_check_master.c
```diff
@@ -230,6 +230,8 @@ static master_test_t master_tests[] = {
   { "Simple Tests", "test_ck_assert_pstr_ne_with_null", CK_FAILURE, "Assertion 't != s' failed: t == (null), s == (null)" },
   { "Simple Tests", "test_ck_assert_ptr_eq", CK_FAILURE, "Assertion 'x == y' failed: x == 0x1, y == 0x2" },
   { "Simple Tests", "test_ck_assert_ptr_ne", CK_FAILURE, "Assertion 'x != z' failed: x == 0x1, z == 0x1" },
+  { "Simple Tests", "test_ck_assert_ptr_null", CK_FAILURE, "Assertion 'x == NULL' failed: x == 0x1" },
+  { "Simple Tests", "test_ck_assert_ptr_nonnull", CK_FAILURE, "Assertion 'x != NULL' failed: x == 0" },
```
9fcc: I thought about it. This assert belongs to `prt_*` macros. And they just print pointer value. So I think printing NULL has no sense. Another way all `ptr_*` macros could be changed to print `NULL` instead of zero. Anyway, it could be a separate commit.

doc/check.texi
9fcc: Sorry, I did it again. I would like to know how do I make such mistakes. Moreover, I've mixed up descriptions.

If checks in my commits fail, you can just wait until I fix everything. I make changes in a hurry, but I can fix them all only when I get home. So I see what's wrong but just can't send fixes.

brarcher:
Looks good. Thanks!

PR #91, opened 2016-12-15
Add regular expressions support for internal tests checking

9fcc:
At the moment it uses POSIX regular expressions. I do not know if windows has those, so if not, I'll change them to pcre. Tests will show.

### DIFF
--- a/tests/check_check_master.c
+++ b/tests/check_check_master.c
@@ -23,6 +23,9 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#if ENABLE_REGEX
+# include <regex.h>
+#endif
 #include <check.h>
 #include <assert.h>
 #include "check_check.h"
@@ -38,12 +41,22 @@ char * test_names_file_name = NULL;
 FILE * line_num_failures = NULL;
 char * line_num_failures_file_name = NULL;
 
+enum ck_test_msg_type_t {
+#if ENABLE_REGEX
+  // For tests with different output on different platforms
+  CK_MSG_REGEXP,
+#endif
+  // Simple text
+  CK_MSG_TEXT
+};
+
 #define MAXSTR 300
 
 typedef struct {
   const char *tcname;
   const char *test_name;
   int failure_type;
+  enum ck_test_msg_type_t msg_type;
   const char *msg;
 } master_test_t;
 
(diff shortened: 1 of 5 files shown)


brarcher:
Jenkins: ok to test

9fcc:
I've added pcreposix support for Windows. CMake build now checks if regular expressions are available, and if not than previous tests are used. At second stage I'll try adding regex and pcre support to autotools build. And at third stage I'll figure out how to add pcre library to AppVeyor.

It's my first cross-platform experience, so cmake files need to be reviewed carefully.

9fcc:
I've figured out how to add regex support to autoconf. But I have a question about AppVeyor. I read that there is a web interface, so can you add PCRE library from there? Or it must be scripted? And can I write script not for powershell, but for bash?

9fcc:
I've decided to split changes into two merge requests. PCRE support will be in the next. And this one could be reviewed and merged.

POSIX systems (Linux and macOS) now have regular expressions support in internal tests. But Windows does not have.

brarcher:
> UPDATE: CMake builds do not print logs on success. So I cannot check if regexp were used for tests. Could Travis configuration be modified to show logs always?

I think the following change will do this for the Travis-CI builds: https://github.com/libcheck/check/pull/95
If it does, I'll merge it in.

brarcher:
> I've figured out how to add regex support to autoconf. But I have a question about AppVeyor. I read that there is a web interface, so can you add PCRE library from there? Or it must be scripted anyway? 

I'm not sure about the web interface. But, here is an example showing how one would script the installation of PCRE:

https://github.com/swig/swig/blob/master/appveyor.yml

If we go that route, we may still want to support Windows without PCRE instead of making it a requirement for Check.

9fcc:
> If we go that route, we may still want to support Windows without PCRE instead of making it a requirement for Check.

Of course it will still be able to compile without PCRE. I can add options to turn PCRE on and off. So it can be tested with or without. All the difference is that without PCRE not all the tests would run. And somebody may easily add regex macros in future (for example to test random number generators).

tests/check_check_master.c
```diff
@@ -455,23 +470,61 @@ START_TEST(test_check_failure_msgs)
...
+        msg_type_str = "";
+      }
 
-    if (strcmp(got_msg, expected_msg) != 0) {      
-      char *emsg;
-      char *tmp = (char *)malloc(MAXSTR);
-      snprintf(tmp, MAXSTR,"For test %d:%s:%s Expected %s, got %s",
-               i, master_tests[i].tcname, master_tests[i].test_name,
-			   expected_msg, got_msg);
+      char _emsg[MAXSTR];
```
brarcher: small thing, but can you rename _emsg to something else? A leading '_' implies that it is internal or something else, when really you mean 'non escaped string'.

tests/CMakeLists.txt
```diff
@@ -72,16 +77,16 @@ set(CHECK_CHECK_EXPORT_HEADERS check_check.h)
 add_executable(check_check_export
   ${CHECK_CHECK_EXPORT_HEADERS}
   ${CHECK_CHECK_EXPORT_SOURCES})
-target_link_libraries(check_check_export check compat)
```
brarcher: Cleaning this up is ok. Can it be in a separate commit, though, as it is not directly related to adding support for regex?

tests/check_check_master.c
```diff
@@ -583,40 +636,79 @@ START_TEST(test_check_test_names)
...
+                  i, master_tests[i].tcname);
+      }
 
-	  free(test_name);
+      free(test_name);
   }
 }
 END_TEST
 
 START_TEST(test_check_all_msgs)
```
brarcher: It did not occur to me that there is a separate test for checking only error messages, then a test for checking all messages (which also include failure messages). Any reason why the failure message only test should simply be removed, as the all messages test should provide equivalent coverage?

brarcher:
I've added support for AppVeyor to build for Cygwin and MinGW. The unit tests do not run yet, but at least there is now coverage for building. I'm looking into getting the unit tests running next.

FYI.

tests/check_check_master.c
```diff
@@ -455,23 +470,61 @@ START_TEST(test_check_failure_msgs)
     tr = tr_fail_array[i - passed];
     ck_assert_msg(tr != NULL, NULL);
     got_msg = tr_msg(tr);
-    expected_msg = master_tests[i].msg;
+    expected_msg = master_test->msg;
+
+    unsigned char not_equal = 0;
+
+    switch (master_test->msg_type) {
```
9fcc: No way. Regular expressions have their own syntax. Regular expression and string are different. Some symbols need to be escaped. Moreover, I think forward  and someday there could be one more `msg_type` for template substitution. It could be useful if some data is platform dependant and could be determined at the initialization. For example it could be data returned by the `localeconv` function.

brarcher:
The MinGW and Cygwin tests are now enabled and running on AppVeyor. I've updated this merge request to fix a merge conflict. Hope you do not mind.

I'm OK with the change. When it passes this round of tests I'll merge it in.

9fcc:
I do rebasing to get merge requests up-to-date, but I can't do it often. So updating is ok, since it results only in one more merge request.

Issue #86, opened 2016-12-13
CMake installation cannot find headers when building binaries on a separate directory


khaotik:
My OS is Ubuntu 16.04. I made a directory `check/Build` to build binaries in `cmake-gui`

```
$ cmake-gui .
$ cd Build/
$ make -j4
[  2%] Building C object lib/CMakeFiles/compat.dir/libcompat.c.o
[  5%] Building C object src/CMakeFiles/check.dir/check.c.o
< ... >
[100%] Linking C executable check_check
[100%] Built target check_check
```
Compilation had no issue, however when I try to install:

```
sudo make install
[sudo] password for khaotik: 
[  5%] Built target compat
[ 30%] Built target check
[ 35%] Built target check_nofork_teardown
[ 51%] Built target check_check_export
[ 89%] Built target check_check
[ 94%] Built target check_nofork
[100%] Built target ex_output
Install the project...
-- Install configuration: "Debug"
-- Installing: /usr/local/include/check_stdint.h
-- Installing: /usr/local/lib/libcompat.a
CMake Error at lib/cmake_install.cmake:40 (file):
  file INSTALL cannot find "/home/khaotik/Build/check/Build/lib/libcompat.h".
Call Stack (most recent call first):
  cmake_install.cmake:41 (include)
  


Makefile:71: recipe for target 'install' failed
make: *** [install] Error 1
```

I fixed the problem by copying the needed file, however I don't think the error should be there on the first place.

brarcher:
Jenkins: ok to test

brarcher:
This issue should now be resolved, so I'll close the issue. If it is not resolved in the next release or in master, let me know.

PR #94, opened 2016-12-18
Fix missing libcompat.h error while doing make install command

9fcc:
Related to issue #86.


brarcher:
Jenkins: add to whitelist

brarcher:
Jenkins: test this please

PR #95, opened 2016-12-18
Always emit unit test output for CMake on Travis-CI

brarcher:
Internally "make test" invokes ctest, and passing the -V flag will print the stdout from the unit tests.


brarcher:
Jenkins: test this please

PR #105, opened 2016-12-24
Install PCRE for MinGW32 build to test regex Check tests

brarcher:


brarcher:
@9fcc This change installs PCRE for the [MinGW32 build](https://ci.appveyor.com/project/brarcher46583/check/build/1.0.447/job/l9j4mxdf31b6c4k3#L15). It does not seem that CMake detects regex support, however, as the ```test_ck_assert_float_nonnan``` test reports Passed. Is something else necessary to use PCRE? For example, the following project finds and uses PCRE using the pcre.h header instead of regex.h:

https://github.com/LuaDist/pcre/blob/master/FindPCRE.cmake

9fcc:
As I said PCRE support will be in a separate merge request. I've almost prepared it. Regex.h is just POSIX standard and does not need any external libraries.
[http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/regex.h.html](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/regex.h.html)
PCRE has the same realization named pcreposix implemented as a separate library. It contains wrapper functions under PCRE functions. So it implements POSIX API and sutable for Windows.

PR #108, opened 2016-12-30
Experiment with ck_assert_int_ge/le and arrays

brarcher:
This is to run against AppVeyor, not actually to commit.


brarcher:
The issue this was experiment with is fixed in Check 0.11.0.

Issue #110, opened 2017-01-10
Problem using check 0.11.0 on OS X


njh:
Hello,

I have just tried using check 0.11.0 on OS 10.12.1:

    Apple LLVM version 8.0.0 (clang-800.0.42.1)
    Target: x86_64-apple-darwin16.1.0

The error message from clang is:

```
/usr/local/Cellar/check/0.11.0/include/check.h:686:37: error: invalid suffix on literal; C++11 requires a space
      between literal and identifier [-Wreserved-user-defined-literal]
  "Assertion '%s' failed: %s == %.*"TM"g, %s == %.*"TM"g", \
                                    ^
                                     
/usr/local/Cellar/check/0.11.0/include/check.h:686:53: error: invalid suffix on literal; C++11 requires a space
      between literal and identifier [-Wreserved-user-defined-literal]
  "Assertion '%s' failed: %s == %.*"TM"g, %s == %.*"TM"g", \
                                                    ^
                                                     
/usr/local/Cellar/check/0.11.0/include/check.h:698:39: error: invalid suffix on literal; C++11 requires a space
      between literal and identifier [-Wreserved-user-defined-literal]
    "Assertion '%s' failed: %s == %.*"TM"g", \
                                      ^
```

I fixed it very easily by putting a space on either side of TM but I am confused as to why nobody else has seen this and why it hasn't been picked up my Travis. Perhaps because I am using a newer version of clang?

I am happy to send a Pull Request, if you think it is the right thing to do.

nick.

brarcher:
Are you using _clang++_ when compiling or _clang_? Travis-CI is setup to use clang, but I wonder if clang++ is being more picky.

It may be that Travis-CI is not using the latest version of Clang, and maybe that is the reason you are hitting this. There is a [test](https://check.ci.cloudbees.com/job/check-linux-github/) run on Jenkins to check that clang++ also works, and that test also did not fail. Again, that may not be the newest version of Clang.

Feel free to submit a pull request to resolve it.

njh:
Thanks Branden. Makes sense.

I wonder if there should be any C++ compiler test in the ```check``` test-suite?
But I guess there are a lot of different permutations (different compilers, different coding standards, different OSs...)

I have double checked and then created a Pull Request in #111.

brarcher:
> I wonder if there should be any C++ compiler test in the check test-suite?

[...]

PR #111, opened 2017-01-15
Fix for "invalid suffix on literal" error in clang

njh:
I get the following error from clang when compiling against check.h with C++:

    /usr/local/include/check.h:760:55: error: invalid suffix on literal; C++11 requires a space between
          literal and identifier [-Wreserved-user-defined-literal]
        "Assertion '%s' failed: %s == %.*"TM"g, %s == %.*"TM"g, %s == %.*"TM"g", \

Adding a space on either side of TM fixes it.

See also #110


brarcher:
Jenkins: ok to test

PR #112, opened 2017-01-16
Also try compiling with c++ compilers

brarcher:


brarcher:
This did not detect the issue reported in https://github.com/libcheck/check/pull/111, so closing request.

PR #114, opened 2017-02-16
Add support to disable building test for check

zhuhaow:
If check is not installed but instead checked into the project as submodule and integrated with CMake then the test will be built and tested with the parent project, which is almost always not what is wanted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,10 @@ set(MEMORY_LEAKING_TESTS_ENABLED 1)
 set(CMAKE_BUILD_TYPE Debug)
 
 ###############################################################################
+# Option
+option(CHECK_DISABLE_TEST "Disable the compilation of test of check itself" OFF)
+
+###############################################################################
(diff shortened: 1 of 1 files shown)


brarcher:
Jenkins: ok to test
^ necessary to start the remaining tests.

brarcher:
This sounds like a fairly typical use case. Thanks for the contribution.

It is a small thing, but could the option be a positive setting instead of an negative setting? Namely:
```
option(CHECK_ENABLE_TESTS "Enable the compilation and running of Check's unit tests" ON)
```
That way, when one disables it one does not need to think about a double negative.

zhuhaow:
Sure.

I was following the option used in libevent, which gives a more clear intention. On the other hand, changing it to affirmative form explicitly defines the default behavior of the building setting.

brarcher:
Jenkins: test this please
(not sure why the status did not update when the build completed, trying again).

brarcher:
Thanks for the contribution!

PR #115, opened 2017-02-19
add definition for m library

DengkeDu:
Like rt library, we should add "ADD_DEFINITIONS(-DHAVE_LIBM=1)".

Signed-off-by: DengkeDu <pinganddu90@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,6 +299,7 @@ unset(CMAKE_EXTRA_INCLUDE_FILES)
 check_library_exists(m floor "" HAVE_LIBM)
 if (HAVE_LIBM)
     set (LIBM "m")
+    ADD_DEFINITIONS(-DHAVE_LIBM=1)
 endif (HAVE_LIBM)
 
 check_library_exists(rt clock_gettime "" HAVE_LIBRT)


brarcher:
I believe that HAVE_LIBRT is used to make a compile time decision about a feature ([here](https://github.com/libcheck/check/blob/0098b4e9a45ed128d8066c504d886b40b5cb717c/lib/libcompat.h#L159) and [here](https://github.com/libcheck/check/blob/b2f4662a99f311e505d1bf101f7777c46300f399/src/check.c#L613)), as well as determine what tests to run (example [here](https://github.com/libcheck/check/blob/82ed3c1919d0bc231f6036a582a56bfd64fa597e/tests/check_check_master.c#L293)). ENABLE_SUBUNIT does something similar. There are currently no compile time decisions made on the existence or absence of libm, which is why HAVE_LIBM is not passed. If you believe that there needs to be a compile time decision, please let me know. Otherwise, I am not sure this change is functionally needed. Let me know if you disagree.

DengkeDu:
Thanks, when I build librepo which depends on the check, the building failed, because the check can't find type  "floor". Using autotools is ok, I can find "-lm" linking options in Makefile.

brarcher:
The CMake check for libm should be the [following](https://github.com/libcheck/check/blob/24e30671819b7d4e7d194373217d69d3a8672904/CMakeLists.txt):
```
check_library_exists(m floor "" HAVE_LIBM)
if (HAVE_LIBM)
    set (LIBM "m")
endif (HAVE_LIBM)
```
If libm has the floor function then libm should be linked against. I do not that that
```
ADD_DEFINITIONS(-DHAVE_LIBM=1)
```
would affect that one way or the other.

If you are linking against libcheck and find that there is a dependences which is not being resolved, are you building libcheck statically and not using the needed flags when linking against it? If so, take a look at the following email thread discussing linking against a statically build libcheck:

https://sourceforge.net/p/check/mailman/message/34265575/

Let me know if that resolves your issue.

DengkeDu:
OK, thanks, it works!

PR #116, opened 2017-03-04
Very minor improvements in C source code

RyDroid:


src/check_log.c
```diff
@@ -94,11 +94,6 @@ void srunner_set_tap(SRunner * sr, const char *fname)
     sr->tap_fname = fname;
 }
 
-int srunner_has_tap(SRunner * sr)
```
RyDroid: It used srunner_tap_fname that is not defined before.

src/check_msg.c
```diff
@@ -24,6 +24,7 @@
 #include <stdlib.h>
 #include <fcntl.h>
 #include <stdio.h>
+#include <unistd.h>
```
RyDroid: For unlink

brarcher: This header is not available if in the Windows environment. Because of this, the MSVC and Visual Studio build fails: https://ci.appveyor.com/project/brarcher46583/check/build/1.0.467.

src/check_log.c
brarcher: Maybe make this a separate commit, instead of combining multiple unrelated improvements in one commit.

RyDroid: @brarcher https://github.com/libcheck/check/pull/119

RyDroid:
I removed from this pull request what I have split in an other one. https://github.com/libcheck/check/pull/119

PR #100, opened 2016-12-20
Adding tests with KWStyle

RyDroid:
I made the file for KWStyle fit to the existing code. However, it could be better to be stricter and change the code in consequence.

PR #117, opened 2017-03-04
Fix errors reported by KWStyle in tests folder

RyDroid:
https://github.com/libcheck/check/pull/100


tests/check_check_log_internal.c
```diff
@@ -72,4 +72,3 @@ Suite *make_log_internal_suite(void)
   
   return s;
 }
-
```
brarcher: I like the extra line at the end of files, though. And, technically all non-empty source files must end in a new line according to the C standard: http://gcc.gnu.org/ml/gcc/2003-11/msg01568.html

tests/check_check_master.c
```diff
@@ -91,194 +93,356 @@ static master_test_t master_tests[] = {
   { "Simple Tests", "test_ck_abort", CK_FAILURE, CK_MSG_TEXT, "Failed" },
   { "Simple Tests", "test_ck_abort_msg", CK_FAILURE, CK_MSG_TEXT, "Failure expected" },
   { "Simple Tests", "test_ck_abort_msg_null", CK_FAILURE, CK_MSG_TEXT, "Failed" },
-  { "Simple Tests", "test_ck_assert", CK_FAILURE, CK_MSG_TEXT, "Assertion 'x == y' failed" },
```
brarcher: I know that these are really long lines, but I do prefer each entry to be on one line. At least for this table, that is. Some of the other long line examples that you have edited are fine.

tests/check_check_master.c
```diff
@@ -611,8 +779,8 @@ START_TEST(test_check_failure_ftypes)
     tr = tr_fail_array[i - passed];
     ck_assert_msg(tr != NULL, NULL);
     ck_assert_msg(master_tests[i].failure_type == tr_rtype(tr),
-                "Failure type wrong for test %d:%s:%s",
```
brarcher: I'm confused what is going on here. Are the tabs being replaced with spaces, or only some of the tabs?

tests/check_check_log_internal.c
RyDroid: Currently (without my commit) there are 2 empty lines. The view of GitHub hides the last one. Check the file with `cat`.

brarcher: Oh, ok.

tests/check_check_master.c
RyDroid: GNU Emacs (default) auto-indentation, I will replace tabs by spaces.

PR #118, opened 2017-03-04
Adding "rules" to .gitignore related to CMake

RyDroid:


.gitignore
```diff
@@ -9,6 +9,8 @@
 *.save
 *.autosav
 *.autosave
+*.swp
```
RyDroid: Created by vim https://stackoverflow.com/questions/821902/disabling-swap-files-creation-in-vim

brarcher:
Jenkins: ok to test

PR #119, opened 2017-03-06
Removing some useless else

RyDroid:


src/check_list.c
```diff
@@ -90,8 +90,7 @@ int check_list_at_end(List * lp)
 {
     if(lp->current == -1)
         return 1;
-    else
-        return (lp->current > lp->last);
+    return (lp->current > lp->last);
```
zapashcanon:
```c
return lp->current == -1 ? 1 : lp->current > lp->last;
```

If you want to reduce it even more...

brarcher:
Jenkins: ok to test

src/check_list.c
RyDroid: I do not remember to have seen ternary expression in the source code, so it may be intentional as it could be considered less clear. Moreover, it is also not supported by some weird compilers, or less well, especially the ones for embed systems. http://embeddedgurus.com/stack-overflow/2009/02/efficient-c-tips-6-dont-use-the-ternary-operator/

Issue #78, opened 2016-11-29
Definition of pid_t not found on plain Windows


kaoh:
`check.h` defines:

    CK_DLL_EXP pid_t CK_EXPORT check_fork(void);

A fix would be:

```
#ifdef WIN32
#define pid_t int
#endif
```

Artox:
This is still a thing!

brarcher:
Take a look at https://github.com/libcheck/check/issues/78 and see if it is what you are thinking, or if you suggest a different fix.

Artox:
Yes thats exactly what I'd like to see.

brarcher:
This should be resolved in the next release of Check.

PR #124, opened 2017-07-23
Define pid_t in check.h for Windows

brarcher:
pid_t is not defined for Windows, so check.h cannot be included directly. To avoid this, define pid_t in check.h for Windows.

To test that this is working, removing the libcompat.h include in ex_output.c, which only uses external API calls and should not need anything from libcomapt.h.

https://github.com/libcheck/check/issues/78

### DIFF
--- a/src/check.h.in
+++ b/src/check.h.in
@@ -63,6 +63,11 @@ CK_CPPSTART
 
 #include <sys/types.h>
 
+#if _MSC_VER
+/* define pid_t for Windows, as it is needed later */
+#define pid_t int
+#endif /* _MSC_VER */
+
 /*
  * Used to create the linker script for hiding lib-local symbols. Shall
  * be put directly in front of the exported symbol.
--- a/tests/ex_output.c
+++ b/tests/ex_output.c
@@ -18,7 +18,7 @@
  * MA 02110-1301, USA.
  */
 
-#include "../lib/libcompat.h"
+
 
 #include <stdlib.h>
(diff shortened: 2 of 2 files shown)


Artox:
If I may comment, checking on _MSC_VER only covers msvc, but what ifsomebody uses mingw (NOT cygwin). Shouldn't something like #ifdef WIN32 && !defined(__CYGWIN__) be more complete?

Artox:
And it can be removed from https://github.com/libcheck/check/blob/master/CMakeLists.txt#L280

brarcher:
Jenkins: test this please

brarcher:
> checking on _MSC_VER only covers msvc, but what ifsomebody uses mingw (NOT cygwin).

I believe that MinGW, Cygwin, MSYS all provide headers that define pid_t. The reason I say this is when libcompat.h was removed from the ex_output.c test program, which would remove the pid_t definition used in the rest of Check, these three platforms continued to build whereas MSVC and VS failed. Here is a link to the Windows-related builds:
https://ci.appveyor.com/project/brarcher46583/check/build/1.0.501

Artox:
That is confusing. Check actually built for me with VS. The only time when I get in trouble is when building another application against Check. Because then pid_t is not defined.
Especially if you consider that CMake actually defines pid_t on a Check build, so it will always be defined no matter what headers do.

brarcher:
This should be resolved in the next release. Thanks for following-up on the bug.

PR #126, opened 2017-08-26
Fix invalid prototype for tcase_name(void)

cleishm:
The existing prototype in check.h is `const char* tcase_name()`. This causes issues on platforms that use strict prototype checking, e.g. the latest clang compiler with -Wstrict-prototypes ('this function declaration is not a prototype [-Werror,-Wstrict-prototypes]').

Changing to `const char* tcase_name(void)` resolves this issue.

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -349,7 +349,7 @@ void tcase_fn_start(const char *fname, const char *file,
     current_test_name = fname;
 }
 
-const char* tcase_name()
+const char* tcase_name(void)
 {
 	return current_test_name;
 }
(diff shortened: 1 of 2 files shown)


brarcher:
Jenkins: ok to test
^ starts the remaining tests

brarcher:
Thanks for the fix. If you are not in the [AUTHORS](https://github.com/libcheck/check/blob/master/AUTHORS) file feel free to submit a request adding yourself.

cleishm:
Thanks for merging @brarcher.

I'll refrain from modifying AUTHORS. I figure it takes more than a single line change to claim any authorship over this great tool. :)

Issue #127, opened 2017-08-26
Make CK_MAX_MSG_SIZE dynamically configurable?


orenbenkiki:
I realize that placing a limit on the message size is a good thing (4GB, shudder...). However, making it a compile-time `#define` seems overly restrictive. How about making it dynamically configurable (keeping 8K as the default limit)?

I just run into a case where I dump two objects to textual format and compare the dumps. The dumps were 2056 characters long... which caused libcheck to die horribly. If the maximal size was configurable I could just bump it up to, say, 16K instead of 8K and solve my problem (my computer has 16GB, surely 16K in a message isn't an issue...?).

I link with the libcheck version I got from apt-get. To modify the limit I'd have to switch to a private patched version - and force my users to do the same. To obey the limit, I'd have to write convoluted code that somehow splits the dumped data on <2K chunks (say, on a line boundary...).

A possible objection is that people might abuse such a dynamic size configuration. True... but if someone _explicitly_ sets the maximal size to 4GB, they can't really complain about the unpleasant side effects - especially if the documentation explained that excessive message sizes are inadvisable.

brarcher:
I've no objection to allowing this to be configurable, perhaps via environment variable. The messages are stored in a file until the test completes, in which case they are read back out. So, only one message should be in memory at a time anyway.

If you would like to prototype making the max size configurable, I would accept the resulting pull request if you are satisfied with the result.

PR #128, opened 2017-08-27
Allow dynamic configuration of maximal assertion message size.

orenbenkiki:
Resolve issue #127.

The new function is annotated as `@since 0.12.0` under the assumption that would be the next released version.


src/check_pack.c
```diff
@@ -38,16 +38,33 @@
...
  * continuously reallocate) one big chunk for the whole message stream.
  * Problems were seen in the wild with up to 4 GB reallocations.
  */
 
+void check_set_max_msg_size(int max_msg_size) {
+    ck_max_msg_size = max_msg_size;
+}
+
+static int get_max_msg_size(void) {
+    if (ck_max_msg_size <= 0) {
```
brarcher: To be consistent with the behavior of the other environment variables, the order of precedence should be:
1) environment variable
2) check_set_max_msg_size()
3) default value

See CK_DEFAULT_TIMEOUT as an example of this.

src/check_pack.c
```diff
@@ -38,16 +38,33 @@
...
+        char *env = getenv("CK_MAX_MSG_SIZE");
+        if (env)
+            ck_max_msg_size = atoi(env);
+
+        if (ck_max_msg_size <= 0)
+            ck_max_msg_size = 4096;
```
brarcher: Could 4096 still be a variable or ```#define``` above, so it is not a magic number?

brarcher:
I think it would be valuable to document this new API call in the manual, maybe somewhere in [this](https://github.com/libcheck/check/blob/master/doc/check.texi#L988) section. Also mention that the environment variable can override the value.

There is an section which shows all the environment variables [here](https://github.com/libcheck/check/blob/master/doc/check.texi#L2250). Also mention the new environment variable there.

Finally, I think it would be valuable to have a new test for this, which shows that if the value is set too low Check does cause an error but if it is set high enough Check continues with the test. Many unit tests in Check run using Check itself, which verifies that certain errors occur when they are expected. The tests themselves are defined up to [this](https://github.com/libcheck/check/blob/master/tests/check_check_sub.c#L2706) point, they are added to the test suite (with whatever environment variables are used) [here](https://github.com/libcheck/check/blob/master/tests/check_check_sub.c#L2708), and the table which says what the output of the tests should be is [here](https://github.com/libcheck/check/blob/master/tests/check_check_master.c#L68). Try adding a test or so to verify the behavior of the new API and environment variable.

orenbenkiki:
Thanks for the review pointers. I fixed the code to give precedence to the environment variable, and updated the TexInfo documentation.

I have difficulty with creating a test though. A too-long message invokes `eprintf`, which writes to `stderr`. It seems the expected-output mechanism doesn't capture such output... is there an example of a test that captures this sort of output I could use as a guide?

doc/check.texi
```diff
@@ -993,6 +993,17 @@ repetitive code that is hard to read.  For your convenience Check
...
 conditions.
 
+@findex check_set_max_msg_size
+@vindex CK_MAX_MSG_SIZE
+The typical size of an assertion message is less than 80 bytes.
+However, some of the functions listed below can generate very large messages
+(up to 4GB allocations were seen in the wild).
+To prevent this, a limit is placed on the assertion message size.
+This limit is 4K bytes by default.
+It can be modified by setting the @code(CK_MAX_MSG_SIZE) environment variable,
```
brarcher: I think that you may need ```{}``` around ```@code```.

brarcher: That is:
```@code{CK_MAX_MSG_SIZE}```

**Comment by orenbenkiki** [comment]
Working on the tests; I see test/check_mem_leaks in the `tests/Makefile.am`, so I can add the new tests there. I don't see them listed anywhere else though, in particular in any of the `CMakeLists.txt` file. Is adding the new tests to `tests/Makefile.am` sufficient?

File: tests/check_env_max_msg_size.c [hunk 4]
```diff
@@ -0,0 +1,70 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+/**
+ * The purpose of this test is to be used by valgrind to check for
```
brarcher: May need to update this comment

tests/check_set_max_msg_size.c
```diff
@@ -0,0 +1,70 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+/**
+ * The purpose of this test is to be used by valgrind to check for
```
brarcher: Same here, the comment needs to be updated

tests/check_set_max_msg_size.c
```diff
@@ -0,0 +1,70 @@
...
+int main (void)
+{
+    int n;
+    SRunner *sr;
+
+    /*
+     * Run the test suite. This is intended to trigger the "Message is too long" error.
+     * Actual success/failure is determined by examining the output.
+     */
+    check_set_max_msg_size(32); // This will have an effect due to no environment variable.
```
brarcher: The two test programs could be collapsed by passing in a value over the command line, as I think that is the only difference, correct?

tests/test_env_max_msg_size.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
brarcher: The two test scripts could be collapsed, and simply run the unit test executable multiple times and check the result.

doc/check.texi
```diff
@@ -993,6 +993,17 @@ repetitive code that is hard to read.  For your convenience Check
 provides a set of functions (actually macros) for testing often used
 conditions.
 
+@findex check_set_max_msg_size
```
brarcher: I think this could be moved to the end of the section, as it would not be expected to be called as frequently as the others. Or, do you disagree?

brarcher:
Jenkins: ok to test
^ will start the remaining tests

brarcher:
> Strange that it seems that cmake doesn't run the memory leak check though...?

The valgrind test is specific to Linux. The original intention of CMake was to support Windows targets, so it did not make much sense to add it. It turns out that some users on Linux do want to use CMake as well. 

Regardless, the unit tests are really to help protect the quality of the project. As long as the project still runs the tests on changes, I think being supported in one or the other build system should not be that important.

doc/check.texi
orenbenkiki: I think that placing it at the end would make it "disappear" for someone who would say "ah, a list of all kind of assertions, I'll skip to the next section and come back here if I want to locate a specific assertion function". The other option is to make it an explicit section of its own but that seems an overkill for a single simple function.

orenbenkiki:
Fixed everything (except for moving the documentation paragraph).

orenbenkiki:
Ready to merge - or is there anything else?

src/check_pack.c
```diff
@@ -38,16 +38,36 @@
...
- * is bigger than this (actually we check against CK_MAX_MSG_SIZE/2).
+ * is bigger than this.
  * The usual size for a message is less than 80 bytes.
  * All this is done instead of the previous approach to allocate (actually
  * continuously reallocate) one big chunk for the whole message stream.
  * Problems were seen in the wild with up to 4 GB reallocations.
  */
 
+void check_set_max_msg_size(int max_msg_size) {
+    if (ck_max_msg_size <= 0) {
```
brarcher: Is it that once the value has been set it cannot be changed? The comment in the prototype says this is the intent: "if used... must be called once...". I am fine with the behavior, just curious as to the rationale.

src/check_pack.c
```diff
@@ -38,16 +38,36 @@
...
  * side. When sending messages, we assure that no single message
- * is bigger than this (actually we check against CK_MAX_MSG_SIZE/2).
+ * is bigger than this.
  * The usual size for a message is less than 80 bytes.
  * All this is done instead of the previous approach to allocate (actually
  * continuously reallocate) one big chunk for the whole message stream.
  * Problems were seen in the wild with up to 4 GB reallocations.
  */
 
+void check_set_max_msg_size(int max_msg_size) {
```
brarcher: Small thing, but can you match the style of the { in this file, namely:
```
void check_set_max_msg_size(int max_msg_size()
{
...
```

src/check_pack.c
```diff
@@ -38,16 +38,36 @@
...
+            ck_max_msg_size = atoi(env);
+        if (ck_max_msg_size <= 0)
+            ck_max_msg_size = max_msg_size;
+        if (ck_max_msg_size <= 0)
+            ck_max_msg_size = DEFAULT_MAX_MSG_SIZE;
+    }
+}
+
+static int get_max_msg_size(void) {
+    if (ck_max_msg_size <= 0)
```
brarcher: Why is this necessary? Should not the setter verify that the value is sane?

Is there a reason that ck_max_msg_size must be signed? Is it valid for it to be negative? Could it be unsigned?

src/check_pack.c
orenbenkiki: I was probably too focused on a specific use case. Fixed.

src/check_pack.c
orenbenkiki: Done.

src/check_pack.c
orenbenkiki: Changed type to `size_t`, which is probably the "right" type for a size.

orenbenkiki:
Changed type of argument to `size_t`, allow modifying the size at any point, and fixed the style.

src/check_pack.c
```diff
@@ -51,21 +51,22 @@ static int ck_max_msg_size = 0;
...
+    if (ck_max_msg_size == 0)
+        ck_max_msg_size = DEFAULT_MAX_MSG_SIZE;
 }
 
-static int get_max_msg_size(void) {
-    if (ck_max_msg_size <= 0)
-        check_set_max_msg_size(0);
+static size_t get_max_msg_size(void)
+{
+    if (ck_max_msg_size == 0)      // If check_set_max_msg_size was not called,
```
brarcher: Should not the getter be making the decision as to if the environment variable exists or not, and let the setter only set the value passed? This way, if the environment variable changed since when the setter was called and the getter was called the correct value would be returned? e.g.:
```
void check_set_max_msg_size(size_t max_msg_size)
 {
     ck_max_msg_size = max_msg_size;
 }

void get_max_msg_size()
{
    size_t value = 0;

    char *env = getenv("CK_MAX_MSG_SIZE");
     if (env)
         value = (size_t)strtoul(env, NULL, 10); // Cast in case size_t != unsigned long.
     if(value == 0)
         value = ck_max_msg_size;
     if (value == 0)
         value = DEFAULT_MAX_MSG_SIZE;

   return value;
}
```

orenbenkiki: Changing the environment variable at run time seems strange, but I guess it could happen. Also, I guess performance isn't an issue, as this is only used when an assertion fails...

orenbenkiki:
Fixed.

brarcher:
Thanks for the changeset! I've merge it in.

If you would like, you can submit a merge request adding an entry into the [AUTHORS](https://github.com/libcheck/check/blob/master/AUTHORS) file for yourself.

Issue #125, opened 2017-08-25
Check for Visual Studio 2010 - without stdbool.h


bel2125:
Currently check does not work for Visual Studio 2010 project, because it misses the `stdbool.h` header from the C99 standard. This include was introduced about a year ago in https://github.com/libcheck/check/commit/6c19a8e3feb7b5d305905efefbe2be424a6adde3

Could we just replace this `bool` by an `int`, so older but still common Visual Studio compilers (2010 and even 2012) could use also recent versions of check?

I could prepare a PR if you like.

brarcher:
Hm, sorry that the change broke you on VS 2010. My expectation with using [AppVeyor](https://github.com/libcheck/check/blob/master/appveyor.yml) would be to catch issues like this; maybe the version of VS there is different than 2010.

The function is internal, so there is no problem if the signature is changed. Alternatively, a compile time check could be made for ```stdbool.h``` and if it does not exist provide a ```#define``` for ```true``` and ```false``` in [libcompat.h](https://github.com/libcheck/check/blob/master/lib/libcompat.h) if it does not exist. That would have the benefit of prevent this breakage in the future if someone introduces usage of true/false but is unaware of the unavailability of ```stdbool.h``` in VC 2010.

Either way, I would be glad to accept a pull request with changes.

brarcher:
I see in your project you also added an AppVeyor builds for VC 2010. If you would like to add a build for that in Check's [AppVeyor configuration](https://github.com/libcheck/check/blob/master/appveyor.yml) as well, feel free.

bel2125:
I removed the stdbool dependency (https://github.com/civetweb/check/commit/9c594550554ad05e95a20eb8a3b35c5c3a235b88) and added an AppVeyor configuration (https://github.com/civetweb/check/commit/5eed08f804f76df1a39df707040b684107ca4ba3).

[...]

PR #129, opened 2017-08-28
Visual Studio 2010 support

bel2125:
Support for C89/C90 compilers like Visual Studio 2010 and 2012

See also https://github.com/libcheck/check/issues/125

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,16 +19,33 @@ version: 1.0.{build}
 #    environment configuration    #
 #---------------------------------#
 
+
 environment:
   matrix:
   - platform: msvc
-  - platform: vs2008_64
+  - platform: vs2008
+    arch: x86
+  - platform: vs2010
+    arch: x86
+  - platform: vs2012
+    arch: x86
+  - platform: vs2013
+    arch: x86
+  - platform: vs2015
+    arch: x86
+  - platform: vs2017
+    arch: x86
   - platform: vs2010_64
+    arch: x64
   - platform: vs2012_64
+    arch: x64
   - platform: vs2013_64
+    arch: x64
   - platform: vs2015_64
+    arch: x64
   - platform: vs2017_64
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    arch: x64
   - platform: vs2012_ARM
   - platform: vs2013_ARM
   - platform: vs2015_ARM
@@ -47,6 +64,10 @@ clone_folder: c:\projects\check
 # scripts that are called at very beginning, before repo cloning
 init:
   - git config --global core.autocrlf input
+  - STRING(SUBSTRING %platform% 1 2 compiler_id)
+  - appveyor AddMessage -Category Information "Compiler ID : '%compiler_id%'"
+  - appveyor AddMessage -Category Information "Architecture: '%arch%'"
+
 
 #---------------------------------#
 #       build configuration       #
@@ -63,8 +84,13 @@ before_build:
   - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
   - if %platform%==msvc call "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat"
   - if %platform%==msvc cmake -G "NMake Makefiles" -DCMAKE_INSTALL_PREFIX=%P%
-  - if %platform%==vs2005_64 cmake -G "Visual Studio 8 2005" -DCMAKE_INSTALL_PREFIX=%P%
-  - if %platform%==vs2008_64 cmake -G "Visual Studio 9 2008" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2005 cmake -G "Visual Studio 8 2005" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2008 cmake -G "Visual Studio 9 2008" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2010 cmake -G "Visual Studio 10 2010" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2012 cmake -G "Visual Studio 11 2012" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2013 cmake -G "Visual Studio 12 2013" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2015 cmake -G "Visual Studio 14 2015" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2017 cmake -G "Visual Studio 15 2017" -DCMAKE_INSTALL_PREFIX=%P%
   - if %platform%==vs2010_64 cmake -G "Visual Studio 10 2010 Win64" -DCMAKE_INSTALL_PREFIX=%P%
   - if %platform%==vs2012_64 cmake -G "Visual Studio 11 2012 Win64" -DCMAKE_INSTALL_PREFIX=%P%
   - if %platform%==vs2013_64 cmake -G "Visual Studio 12 2013 Win64" -DCMAKE_INSTALL_PREFIX=%P%
@@ -105,7 +131,7 @@ before_build:
 build_script:
   # build code and unit tests
   - if %platform%==msvc nmake
-  - if %platform%==vs msbuild "ALL_BUILD.vcxproj"
+  - if %compiler_id%==vs msbuild /p:Platform=%arch% "ALL_BUILD.vcxproj"
   - if %platform%==cygwin bash -c "make"
   - if %platform%==mingw32 mingw32-make
   - if %platform%==mingw64msys bash -c "make"
@@ -114,10 +140,11 @@ test_script:
   # run unit tests
   - echo Running unit tests...
   - if %platform%==msvc nmake test VERBOSE=1 CTEST_OUTPUT_ON_FAILURE=TRUE
-  - if %platform%==vs ctest --extra-verbose -C Release
+  - if %compiler_id%==vs ctest --extra-verbose -C Release
   - if %platform%==cygwin bash -c "make check"
   - if %platform%==mingw32 tests\check_check.exe
   - if %platform%==mingw64msys bash -c "tests/check_check"
 
 on_finish:
   - if %platform%==cygwin bash -c "cat tests/test-suite.log || true"
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,6 @@ version: 1.0.{build}
 environment:
   matrix:
   - platform: msvc
-  - platform: vs2005_64
   - platform: vs2008_64
   - platform: vs2010_64
   - platform: vs2012_64
@@ -85,6 +84,7 @@ before_build:
   - if %platform%==mingw64msys bash -c "./configure"
 
 #Platforms supported by AppVeyor, as of 2017-08-28
+#see https://www.appveyor.com/docs/build-environment/
 #Generators
 #  Visual Studio 15 2017 [arch] = Generates Visual Studio 2017 project files.
 #                                 Optional [arch] can be "Win64" or "ARM".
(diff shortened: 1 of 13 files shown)


bel2125:
I added about every version of Visual Studio, that *in theory* should be supported. *In practice* some builds do not work, since the compiler is missing on the build image.
Visual Studio 2010 and 2012 work now:
```
Environment: platform=msvc     ... OK
Environment: platform=vs2005_64 ... No CMAKE_C_COMPILER could be found
Environment: platform=vs2008_64 ... No CMAKE_C_COMPILER could be found
Environment: platform=vs2010_64 ... OK
Environment: platform=vs2012_64 ... OK
Environment: platform=vs2013_64 ... OK
Environment: platform=vs2015_64 ... OK
Environment: platform=vs2017_64 ... No CMAKE_C_COMPILER could be found.
Environment: platform=vs2012_ARM ... OK
Environment: platform=vs2013_ARM ... OK
Environment: platform=vs2015_ARM ... OK
Environment: platform=vs2017_ARM ... No CMAKE_C_COMPILER could be found.
Environment: platform=cygwin ... OK
Environment: platform=mingw32 ... OK
Environment: platform=mingw64msys ... OK
```
I do not really need 2005 and 2008 any more, and I do not yet need 2017. 

For 2017, I think the appveyor.yml file needs `image: Visual Studio 2017` (see [documentation](https://www.appveyor.com/docs/build-environment/#visual-studio-2008)) to select a different build machine than the standard 2015 build image. I don't know if you need this, or if it makes sense for the "check" project now, or if it has any side effects (like using a slower, experimental, less stable build environment).

I could remove the 2005, 2008 and 2017 builds from the AppVeyor build list, or have a short look at either 2017 and/or the older ones. Just tell me what you prefer.

And please have a look at the changes, if you are willing to merge this, or I should do something different, in particular with the INFINITE and NAN issues.

bel2125:
Visual Studio 2017 and 2008 are fixed now: https://ci.appveyor.com/project/brarcher46583/check/build/1.0.514
 - only the Visual Studio 2017 build needs to move to the new 2017 image, while the rest may still use the standard (operationally well-tried) 2015 image
- Visual Studio 2008 uses a different syntax

Only 2005 is still failing, but it seems AppVeyor dropped all versions prior to VS 2008 (https://www.appveyor.com/docs/build-environment/#visual-studio-2008) - anyway, I thing supporting VS2008 is already more than I'd expected, VS2005 is pretty similar to VS2008, ... I we can't change the AppVeyor images anyway. So I will remove the VS2005 build from appveyor.yml.

bel2125:
Continuous integration using AppVeyor is working now for Visual Studio 2008, 2010, 2012, 2013, 2015 and 2017: https://ci.appveyor.com/project/brarcher46583/check/build/1.0.515
I also added some ARM processor builds for VS 2012, 2013, 2015 and 2017.

It would be possible to do 32 and 64 bit builds and add debug and release builds, effectively 4 builds for a Visual Studio version, but I think this would be another pull request, no longer related to https://github.com/libcheck/check/issues/125

brarcher:
Jenkins: ok to test
^ runs the remaining tests

brarcher:
> Continuous integration using AppVeyor is working now for Visual Studio 2008, 2010, 2012, 2013, 2015 and 2017

Nice!

> It would be possible to do 32 and 64 bit builds and add debug and release builds, effectively 4 builds for a Visual Studio version, but I think this would be another pull request, no longer related to #125

I will defer to you if you think that adding these builds to prevent Check from breaking on them has value to you. I do not normally use Windows (so there are many things that I do not understand about its build environment). However, I do want Check to be operational in different Windows' environments, if possible.

appveyor.yml
```diff
@@ -52,7 +63,17 @@ before_build:
   - set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
   - if %platform%==msvc call "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat"
   - if %platform%==msvc cmake -G "NMake Makefiles" -DCMAKE_INSTALL_PREFIX=%P%
-  - if %platform%==vs cmake -G "Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX=%P%
+  - if %platform%==vs2005_64 cmake -G "Visual Studio 8 2005" -DCMAKE_INSTALL_PREFIX=%P%
```
brarcher: This adds more variations for compilation, but not running the unit tests. Should they also run the build unit tests?

lib/gettimeofday.c
```diff
@@ -29,7 +29,7 @@
 
 int gettimeofday(struct timeval *tv, void *tz)
 {
-#if _MSC_VER
```
brarcher: Out of curiosity, what the difference? Not all versions of _MSC_VER evaluated to 1 or true?

tests/check_check_sub.c
```diff
@@ -267,7 +295,9 @@ START_TEST(test_ck_assert_int_lt)
 {
   int x = 2;
   int y = 3;
+
```
brarcher: What adding the spaces necessary for C89 support? Or is this a whitespace only change?

tests/check_check_sub.c
```diff
@@ -2693,8 +2917,11 @@ void exit_handler ()
 
 START_TEST(test_ignore_exit_handlers)
 {
+  int result;
+
   record_test_name(tcase_name());
-  int result = atexit(exit_handler);
+
+  result = atexit(exit_handler);
```
brarcher: If you did all this by hand, I give you credit for your dedication. (:

tests/check_check_sub.c
```diff
@@ -27,6 +27,17 @@
 #include <check.h>
 #include "check_check.h"
 
+#if !defined(NAN)
```
brarcher: As mentioned earlier, is there a way to put this in libcompat.h?

tests/check_check_sub.c
```diff
@@ -37,6 +37,40 @@ double NAN;
 #if !defined(INFINITY)
 double INFINITY;
 #endif
+#if !defined(isnan)
```
brarcher: Can this be put into the lib dir as a reimplementation of these if they do not exist?

brarcher: See timer_settime as an example.

appveyor.yml
bel2125: Aren't the unit tests run just for the code built by the selected compiler?
At least I saw some tests running.

But there are some extra lines:

if %platform%==vs msbuild "ALL_BUILD.vcxproj"

if %platform%==vs ctest --extra-verbose -C Release

I will check what difference they really make. AppVeyor seemed pretty complete without them ...

lib/gettimeofday.c
bel2125: #if _MSC_VER 
may lead to compile errors - maybe a standard violation for older compilers.
Nevertheless I saw this error (I could try to find the AppVeyor log, if you are really curious).
If you just want to detect if it's defined at all, use
#ifdef _MSC_VER

tests/check_check_sub.c
bel2125: It's not required by the standard. The reason was:

I made all the changes in tests/check_check_sub.c manually, and they were quite a lot.
So I finally came up with a "let's have all functions look like this":

```
START_TEST(test_something) 
{
  type1 var1;
  type2 var2;

  record_test_name(tcase_name());

  test code
  more test code
}
END_TEST
```

Just to get this structure at first sight. This helped me to keep an overview what is already handled and what is not. Maybe it's helpful for maintenance as well.

But it's your project, so choose whatever structure you like.

Do you actually have something like a `clang-format` file for automatic code formatting?

tests/check_check_sub.c
bel2125: Yes I did it by hand. I think, there is no other way to do it.
In the end, it turned out as somewhat more work as initially expected ("just replace stdbool.h").
But it should be in a form that can be maintained in the project, so the test has to work as well, right?

tests/check_check_sub.c
bel2125: The definition can be moved there.
But the definition alone is not enough - there needs to be some code to initialize the value. You need to generate the NAN and INFINITY at runtime, basically at start time of the project. Currently there is no initialization function in libcompat.c.
I can try if this can be done with some preprocessor magic, or by nesting some initializations.
But it's probably not really working if there is no initialization function.

tests/check_check_sub.c
bel2125: I will try (https://github.com/libcheck/check/pull/129/commits/37a1188e9c23fbebec4829c3f105fdf2977f4277)

appveyor.yml
brarcher: I believe the ctest lines are what run the tests. When I looked at the AppVeyor output for the new platforms the unit tests were not being run.

tests/check_check_sub.c
brarcher: I'm fine with the whitespace changes, just curious. I do not have a clang-format file for formatting.

bel2125:
Infinity/NaN code moved, AppVeyor is working (https://ci.appveyor.com/project/brarcher46583/check/build/1.0.517), Travis as well, but I just don't know what the test called "default" is supposed to be.

"Default" indicates a failed test (https://check.ci.cloudbees.com/job/github-merge-builder/284/)

bel2125:
To move INFINITY and NAN into libcompat.h, I would need an initialisation function.
I can't assign this as an initializer in VS 2010: https://ci.appveyor.com/project/brarcher46583/check/build/1.0.519/job/su1wjaku63aukeed#L226

Is there some function in the check library, that is always called first?

brarcher:
> but I just don't know what the test called "default" is supposed to be.

There are some tests which are run on a Jenkins instance. For whatever reason Github calls it "default". The test which is failing is the first one it attempts, [here](https://check.ci.cloudbees.com/job/check-prereleasecheck-github/280/consoleFull). The test runs the ```make prereleasecheck``` target (which builds check, runs the unit tests, then packages Check up in a tarball for release), then unpacks the tarball and build it using CMake. The CMake build is failing due to the following:
```
16:59:33 CMake Error at lib/CMakeLists.txt:68 (add_library):
16:59:33   Cannot find source file:
16:59:33 
16:59:33     fpclassify.c
16:59:33 
16:59:33   Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
16:59:33   .hxx .in .txx
```
I think that fpclassify.c needs to be mentioned in the autotools files so it is packaged when creating a release.

brarcher:
> Is there some function in the check library, that is always called first?

If getting these functions into libcompat is turning into a hassle, I would not worry about it then. The functions are used only in a unit test, and only in one file at that. I was hoping to have all the code for platform specific support in one spot, but having an initialization in Check for test code may be too much.

If you want to keep trying, you could put the initializations in the code which sets up the tests to be run, such as in check_check_sub.c's make_sub_suite(). Alternatively, you could have a getter function in that file which returns the appropriate value of INFINITY or NAN and the value is just computed each time.

bel2125:
```
16:59:33 CMake Error at lib/CMakeLists.txt:68 (add_library):
16:59:33   Cannot find source file:
16:59:33 
16:59:33     fpclassify.c
16:59:33 
16:59:33   Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
16:59:33   .hxx .in .txx
```

I don't know. I already added fpclassify to [lib/CMakeLists.txt:24](https://github.com/civetweb/check/blob/8835f52e1429fb74c3d3615b38e175180e6ae853/lib/CMakeLists.txt#L24), in exactly the same way all other *.c files are added as well. It is also located in the same directory.
I could try if adding a prototype for `fpclassify` will help, but I don't think so. Still I don't have any other idea what might be missing here. I don't know enough detail how the check build works.

Do you know what has to be done?

> If you want to keep trying, you could put the initializations in the code which sets up the tests to be run, such as in check_check_sub.c's make_sub_suite(). Alternatively, you could have a getter function in that file which returns the appropriate value of INFINITY or NAN and the value is just computed each time.

I think I found some combination of preprocessor magic and global variables (`DOUBLE_ZERO`) to make it work without initialization function. The `DOUBLE_ZERO` is actually defined in `fpclassify.c`, so I don't understand how `fpclassify.c` could be missing, since all the AppVeyor and Travis CI builds work now, and almost all tests - there would be a link error if the `DOUBLE_ZERO` symbol is missing.

All tests work now, except for the one in Visual Studio 2008. It seems the build is finally working, but some unit tests are failing: [log starts here](https://ci.appveyor.com/project/brarcher46583/check/build/1.0.529/job/0xqfb9vx7xaoo0xy#L1756)
I don't know the tests well enough to see at the first sight what's going on. Do you want to have a look?
Or should we just drop VS2008 for the moment - in the end I don't really need anything older than VS2010, I just handled VS2008 because "it was there" on AppVeyor.

bel2125:
I've removed the Visual Studio 2008 build and unit test - now all automatic CI test run, including "default".

While the Visual Studio 2008 build works, but the unit test fails - [maybe here](https://ci.appveyor.com/project/brarcher46583/check/build/1.0.531/job/9i530uhw8rtcp3v2#L2330), or maybe this is intended ... I don't know enough how these test logs should look like
Maybe this is worth looking at, but it's not related to #125 (Visual Studio 2010 support).

Do you think this can me merged now?

lib/fpclassify.c
```diff
@@ -0,0 +1,48 @@
+/* Replacement function for isnan/isinf/isfinite for processors using
+ * IEEE 754 floating point format.
+ * Copyright (C) 2017, bel2125
+ * License: LGPL v2 or MIT (as you like).
```
brarcher: Is your intention to provide this file with a different license than that of Check? Check uses LGPLv2.1. I do not think that there is a functional difference between v2 and v2.1, except that the license's name changed in v2.1.

brarcher: Do you mind if this file has the license which is mentioned in the other files, namely:
```
/*
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.
 *
 * This library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 * Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
 * License along with this library; if not, write to the
 * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston,
 * MA 02110-1301, USA.
 */
```

lib/fpclassify.c
```diff
@@ -0,0 +1,48 @@
+/* Replacement function for isnan/isinf/isfinite for processors using
+ * IEEE 754 floating point format.
+ * Copyright (C) 2017, bel2125
```
brarcher: This caught me off guard, as I've not received a contribution which includes its own copyright. However, looking up what this implies it does not really change anything. Because you are not transferring the copyright to another legal entity or person you retain the copyright of your committed changes regardless of adding a header.

You are welcome to include the copyright header, but I'm not sure that it is needed, as the copyright still stays with you. Do you see copyright headers from individual contributors in other open source projects?

brarcher:
The changes themselves look fine and are ready to merge. I've just a question about the license of the fpclasify.c file.

In addition, you are welcome to add an entry for yourself in the [AUTHORS](https://github.com/libcheck/check/blob/master/AUTHORS) file.

bel2125:
If I write open source software, I usually do it with the MIT license. A main difference between MIT license and LGPL is, somewhat simplified: for LGPL, if you modify the source, you must publish this source or provide it in some way for your users. for MIT, you still may do this, but you don't have to.

My brief "LGPL v2 or MIT (as you like)" just means, you can choose your license. Since MIT has less restrictions, you could use MIT licensed files in GPL/LGPL projects - you can not use GPL files in MIT projects (for LGPL you can, if you make them a library). My intention was like "if you use it within check, use it as LGPL" and "if you use only this file, use it as MIT (or LGPL)".

Maybe a more clear way to do this would be:
- I create a separate repository with this file with MIT license,
- and the copy within check should stay with a LGPL license
This way, you have only LGPL code in check, and someone who needs a MIT license can take it from my separate repository.

bel2125:
I set the fpclassify.c here in check to LGPL, so it has the same license as the rest of the code.
I'll keep an MIT version here https://github.com/bel2125/floatmagic

So you have a "LGPL only" code in check, and someone who needs/prefers MIT can have it there.

brarcher:
Thanks for the changeset! I've merged the pull request in. The addition of the Visual Studio tests on AppVeyor should prevent a change from coming in that breaks one of those compilers.

PR #142, opened 2017-11-16
 Typo fix in ck_assert_str_eq documentation

foviedoITBA:


brarcher:
Thanks for finding this!

Could you create a merge request against master as well, so that the documentation distributed with the project is also up to date?

PR #149, opened 2017-12-11
Add overloadable attributes for vsnprintf and snprintf

illiminable:
- Adds overloadable attributes for clang.

### DIFF
--- a/lib/libcompat.h
+++ b/lib/libcompat.h
@@ -234,12 +234,20 @@ CK_DLL_EXP int timer_delete(timer_t timerid);
 #include <stdarg.h>
 
 #if !HAVE_VSNPRINTF
-CK_DLL_EXP int rpl_vsnprintf(char *, size_t, const char *, va_list);
+CK_DLL_EXP int
+#if defined(__clang__)
+__attribute__((overloadable))
+#endif
+rpl_vsnprintf(char *, size_t, const char *, va_list);
 
 #define vsnprintf rpl_vsnprintf
 #endif
 #if !HAVE_SNPRINTF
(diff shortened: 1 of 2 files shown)


illiminable:
Ran into this while trying to port to Chrome OS.

brarcher:
To date the project has been building against GCC and Clang both, but no build issue has been found. Are you seeing an issue on more recent versions of Clang? Can you provide more description to the problem which was found and how this solves it, so I can have a better understanding of what issue is being addressed?

illiminable:
It may only be when using the -verify option, and if the standard library is appropriately annotated.

../../check-0.9.13/lib/libcompat.h:194:16: fatal error: redeclaration of 'rpl_vsnprintf' must have the 'overloadable' attribute
CK_DLL_EXP int rpl_vsnprintf(char *, size_t, const char *, va_list);
               ^
../../../../../../../usr/include/stdio.h:396:12: note: previous overload of function is here
extern int vsnprintf (char *__restrict __s, size_t __maxlen,
           ^
../config.h:362:19: note: expanded from macro 'vsnprintf'
#define vsnprintf rpl_vsnprintf
                  ^

brarcher:
Oh, OK. Is using -verify something that should be done in the CI builds to ensure now new breakage is introduced in the future? It sounds from the [Clang documentation](https://clang.llvm.org/docs/InternalsManual.html#the-diagnosticclient-interface) that -verify is used to pass information from the compiler to an IDE, but it is not entirely clear to me.

Can you mention some of your comment in the commit message? I'll then merge the pull request.

PR #150, opened 2018-01-10
Add make install

ckdarby:
This step is missing from the README


brarcher:
Thanks! I'll let the tests pass (which are not related to the change, but necessary to pass to merge) then merge it in.

brarcher:
Jenkins: ok to test
^ starts the remaining tests

PR #158, opened 2018-06-12
Fix START_TEST to look like valid C code.

filbranden:
Instead of exporting the defined name as a bare function, export a struct that has a pointer to the function, but also its name, file and line number where it is defined.

Store that information into a new `struct TTest`.

After this commit, <code>START_TEST(<i>testname</i>)</code> will create three definitions:
- <code><i>testname</i>_fn</code>: The actual function;
- <code><i>testname</i>_ttest</code>: A `struct TTest` with the information about it;
- <code><i>testname</i></code>: A pointer to <code><i>testname</i>_ttest</code>.

Functions `tcase_add_test()` and friends are updated to take a `TTest *` argument rather than a `TFun` and separate name. The runners are updated to find that information inside the linked `tc->ttest`. The call to `tcase_fn_start()` is moved from the defined functions to the runners (both the "fork" and the "nofork" one) which call it just before invoking the test function.

A nice side-effect is that END_TEST is now optional, though the empty `#define` is kept for backwards compability.

Tested:
- `make check` still passes.
- Removing `END_TEST` from test cases still produces valid code that builds and passes tests.

This PR should address [this TODO item](https://github.com/libcheck/check/blob/0.12.0/TODO#L64).

NOTE: I'm not 100% sure I like the name `TTest` but I wonder what would be a better name for that item... I thought of `TFunc` but that's too similar to `TFun`. What would you call that? A test fixture? A test function? A test method? Clarifying the terminology would be helpful here...

Cheers!
Filipe

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -247,21 +247,21 @@ void suite_add_tcase(Suite * s, TCase * tc)
     check_list_add_end(s->tclst, tc);
 }
 
-void _tcase_add_test(TCase * tc, TFun fn, const char *name, int _signal,
-                     int allowed_exit_value, int start, int end)
+void _tcase_add_test(TCase * tc, const TTest * ttest,
+                     int _signal, int allowed_exit_value,
+                     int start, int end)
 {
     TF *tf;
 
-    if(tc == NULL || fn == NULL || name == NULL)
+    if(tc == NULL || ttest == NULL)
         return;
     tf = (TF *)emalloc(sizeof(TF));   /* freed in tcase_free */
-    tf->fn = fn;
+    tf->ttest = ttest;
     tf->loop_start = start;
     tf->loop_end = end;
     tf->signal = _signal;       /* 0 means no signal expected */
     tf->allowed_exit_value =
       (WEXITSTATUS_MASK & allowed_exit_value);   /* 0 is default successful exit */
-    tf->name = name;
     check_list_add_end(tc->tflst, tf);
 }
 
--- a/src/check.h.in
+++ b/src/check.h.in
@@ -122,6 +122,16 @@ typedef void (*SFun) (void);
 typedef struct Suite Suite;
 
 /**
+ * Type for a test, which wraps a test function
+ */
+typedef struct TTest {
+    const char *name;
+    TFun fn;
+    const char *file;
+    int line;
+} TTest;
+
+/**
  * Creates a test suite with the given name.
  *
  * Create a suite, which will contain test cases. Once
@@ -214,8 +224,8 @@ CK_DLL_EXP void CK_EXPORT tcase_set_tags(TCase * tc,
  *
  * @since 0.9.2
  * */
-#define tcase_add_test_raise_signal(tc,tf,signal) \
-   _tcase_add_test((tc),(tf),"" # tf "",(signal), 0, 0, 1)
+#define tcase_add_test_raise_signal(tc,ttest,signal) \
+   _tcase_add_test((tc),(ttest),(signal), 0, 0, 1)
 
 /**
  * Add a test function with an expected exit value to a test case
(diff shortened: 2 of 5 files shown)


brarcher:
The AppVeyor failures are with the 2010 and 2012 Visual Studio releases not supporting inline initialization of structs. Errors like the following are reported:
```
C:\projects\check\tests\check_check_exit.c(31): error C2059: syntax error : '.' [C:\projects\check\tests\check_check.vcxproj]
```
The versions after that, namely 2013, 2015, and 2017, do not find an issue. Check does not claim it supports specific versions of that compiler, only that it supports it. It may be acceptable to drop support for the older versions.

The other checker, "default" is a Jenkins job which is failing to clone from the pull request. I'll need to look into that.

The change look good to me at first pass. I'll look over it a little more carefully.

filbranden:
> The AppVeyor failures are with the 2010 and 2012 Visual Studio releases not supporting inline initialization of structs.

I imagine they don't support initialization with named fields, but if we do initialization by position that might fix it.

Instead of:

```c
static const TTest __testname ## _ttest = {\
  .name = ""# __testname,\
  .fn = __testname ## _fn,\
  .file = __FILE__,\
  .line = __LINE__,\
};\
```

We would use:

```c
static const TTest __testname ## _ttest = { ""# __testname, __testname ## _fn, __FILE__, __LINE__ };\
```

Which is OK, not that bad to read...

Let me know if you'd like me to push something like that.

Otherwise, we could add a requirement on whatever C standard (C99? C11?) specified the ability to include named fields in struct initialization...

brarcher:
Try pushing a change to initialize by position and see if it builds. If it does, we'll go with it.

filbranden:
> Try pushing a change to initialize by position and see if it builds. If it does, we'll go with it.

Done, let's see how that goes... Thanks!

brarcher:
Hm, it still seems unhappy:
```
check_check_exit.c(31): error C2059: syntax error : '.' [C:\projects\check\tests\check_check.vcxproj]
```
but I do not see that your second push changed the code. Did your second attempt include the change?

filbranden:
I had changed `check.h` instead of `check.h.in`, duh... Fixed now, let's see how it goes.

brarcher:
That make the Visual Studio build much happier.

I think I figured out what broke with the Jenkins job, a change on GitHub's side to deprecate weak crypto that the job was using. That should now be resolved, and I'll rekick the job.

Jenkins: retest this please

brarcher:
Jenkins will fetch now, but the updated Jenkins nodes are missing at least one tool. I'll need to have it installed so the various builds and tests can make it through. I've sent a request to have the tool installed, and when it is I'll rekick.

filbranden:
Sounds good, thanks!

By the way, do you like the name `TTest`? I wasn't too sure about it... Is the right terminology to have TestSuite → TestCase → Test? I thought of "Test Function" but the name `TFun` is already taken (and `TFunc` would be too similar...) Let me know if you'd like any changes in this area...

Cheers!
Filipe

filbranden:
Hi @brarcher, let me know if you still require an action on my part. (For instance, if Jenkins is now fixed, would you like me to touch my tree, pushing a no-op change like a `git commit --amend` with no additional changes, just to trigger it again?)

Thanks!
Filipe

brarcher:
I got a response from CloudBees, who provides the Jenkins instances for free to FOSS projects. They said it would take over a day for an engineer to fix the VMs to install the missing package, and they did not want to do it. In response to this, I'll need to move all the tests off of Jenkins and onto Travis-CI. When I finish that, I'll merge master into the task branch, which will kick off the tests again.

I cannot think of a better term for TTest. Calling it a 'fixture' may be good, such as TestFixture. Though, the name does not mean anything without context.

If you want to change TTest to TestFixture (or another name if one comes to mind), besides that you should have nothing to do on your side if the tests which I'll port to Travis-CI pass. If one of the tests do find an issue, I'll poke you.

brarcher:
I've moved the automated tests to Travis-CI [here](https://github.com/libcheck/check/pull/159), and am updating the branch to pick up these changes.

brarcher:
All the tests passed, the change is good to go. Thanks for the contribution. If you would like to submit a merge request adding yourself to the AUTHORS file, feel free.

ThomasHabets:
Looks like this change makes it impossible to create a test registry, since now the actual function is not predictable. It also doesn't seem possible to be version agnostic about this breaking change.

Do you have a recommendation for how to do [this](https://github.com/ThomasHabets/arping/blob/61b5819bb123bceff66a7718e70989f070053143/src/arping_test.c#L58) in a way that works before and after this change?

ThomasHabets:
Alternatively, do you have a macro I can `#ifdef` on to duplicate my code for before and after this change?

Issue #154, opened 2018-03-10
Consider creating deb as part of releases


wltjr:
If possible it would be nice to have a deb available as part of releases. I was using some newer features of check, and had problems on Travis due to old version of check. I went looking for newer package, and seems one does not exist. Or at least I could not find one.

For some projects of mine I create deb and rpm as part of deployment from Travis. I was wondering if this project could be willing to consider doing the same. I can submit a PR with the initial changes for such.

The idea is to have a package usable for Travis. Since it does not seem an updated deb of check 0.12.0, or even 0.11.0, exist anywhere. If building under Travis it should be usable just the same under Travis. It is not so much for general usage, but that may work just the same.

Thank you for your consideration.

brarcher:
If I understand correctly, you have a project running on Travis and you would like to test that project using Check. However, as the images used on Travis do not have an up-to-date packages available for Check, you cannot use newer Check features. If a deb archive existed you could install that during your Travis builds. Is my summary correct?

Is the issue here that downstream Linux distros are not updating their packages of Check, and as such there is no deb archive? Is providing a deb file from Check useful because it would available in a known location, such as the releases page on GitHub?

wltjr:
Yes your summary is 100% correct. This project [failed on travis](https://travis-ci.org/Obsidian-StudiosInc/asspr/jobs/349607066#L594) using newer features. I plan to use check with other projects. That was more a test, and first feature I used was added in 0.12.0.

Yes the issue is downstreams distros, [ubuntu](https://packages.ubuntu.com/bionic/check) and [debian](https://packages.debian.org/sid/check) do not have updated packages of check. I could not find any third party packages either. Yes it would be useful to be in a known location as a result.

brarcher:
I'm not sure I want to take on maintaining deb packages for releases, sorry. That, and I'm don't think it resolves the root problem. I'm not sure why downstream has not updated the repos in a long time. Perhaps they are still watching Sourceforge and did not start tracking on GitHub. The [Check webpage on Sourceforge](http://check.sourceforge.net/) clearly says to go to GitHub and provides a link. It may be worthwhile to send a note to Ubuntu or Debian and inquire about updating the Check package. Maybe they have a reason for not doing so, though I've not heard anything.

wltjr:
[...]

PR #163, opened 2018-07-06
Modifications to CMakeLists.txt and .travis.yml for auto release to Github

wltjr:
Per #154 I have made initial modifications to CMakeLists.txt for auto creation of binary dep, rpm, and tar.gz. It could also do a [dmg format for OS X](https://cmake.org/cmake/help/latest/module/CPackDMG.html). I have no experience there, but I assume you just add in another generator and any specifics for that package.

I also made initial modifications to .travis.yml for auto release deployment to Github under releases. I added in conditionals so it is only done on tag on the master branch, when under Linux. That would need to be modified if you add in dmg format. Also could add tar.gz to the binaries uploaded to github. Ideally a changelog as well, but I have not got that part automated myself yet.

The cmake changes will work as is. The modifications to .travis.yml require an [additional step to generate a key](https://docs.travis-ci.com/user/deployment/releases/#Authenticating-with-an-OAuth-token). Which I recommend copying the .travis.yml, to _/tmp_ before generating the key. It will re-format the yaml file, which I dislike. You just need to do that for the key. Once you do, replace the file with the one from _tmp_ plus the key from the newly generated file. Or other ways, but that is my preferred workflow. Either way need to replace ```<REPLACE_ME_WITH_SECURE_KEY>``` with a secure key.

Feel free to tweak and modify as needed. This is just initial work to get the ball rolling. The purpose is to have the latest libcheck available for testing other packages within Travis. If they are used outside of such that should be fine. Though that might bring about a larger support issue with official binaries.

Thanks for consideration of this PR and my needs for newer libcheck under Travis.


brarcher:
Travis-CI says it [could not parse](https://travis-ci.org/libcheck/check/requests) the .travis.yml file correctly, and did not build. It does not say what line it failed on.

brarcher:
Oh, maybe it is expecting the secure key. (:

I'll see about generating a key and adding another commit.

wltjr:
I hope it was just the key. I did not think about the yaml parsing of my placeholder. Hopefully that was it and not some accidental tab. The env var can be anywhere or not even a var. I just tend to put it up top so its easy to access for new tags/releases.

That is an odd failure on the MIA readme for some builds. Also the Test Builds was ideally for all builds just to put deploy on its own. For some reason its splitting off a single build under Test builds and all others under Test. Ideally only want to run deploy on tags. Could even add conditionals to not run other stuff on tag. Assuming things were tested pre tag. Usually if something is failing for a test build on a tag, it is a bad tag. I need to update my own stuff to only do builds for non tags/releases. But its semi moot.

Anyway something went wrong with my jobs modification. It should be like before, all test builds under 1 section. This way on release/tag you will have 3.

wltjr:
Looks like my extra stage caused the split, per the [travis docs on build stages](https://docs.travis-ci.com/user/build-stages/matrix-expansion/). Dropping that puts all back into one.

brarcher:
The two jobs which are failing test that autotools can create a release tarball, and that the tarball contains everything necessary to reproduce a release. During the creation of the tarball the README.md file is changed to README. If I recall correctly, this is because the autotools process requires a README and does not recognize a README.md. The proposed changes expect a README.md, but in this case there is only a README.

How important is the README.md file? Could it be made optional?

wltjr:
Readme is semi moot in deb or rpm, short of them being official and others using it. Not sure how many would rely on the readme or even read it.

I recall that from autotools. I used to go with a symlink from README -> README.md. I believe it can be made optional. Though I think it maybe better to see about making the entire cpack portion optional. Like for those failing autotools tests, making its own release via autotools. No need to really have cmake make deb, rpm, etc for those builds. Really only needs to be run with tag/release on deploy. Unless you want to test it for some reason and/or use libcheck via installing deb. I do that for jem and assp, after doing the build, test, install it and run tests in actual env via generated deb.

The README portion optional via tests etc, or the entire cpack section being optional, or symlink to README and then no worries about autotools vs cpack. What ever is preferred.

brarcher:
When one checks out Check what would one need to do to manually build the rpm file? Is that triggered with the CMAKE_BUILD_TYPE=Release variable? What would it mean to make the CPack section optional?

Could just the README.md be dropped from the CPack?

What would I need to do to trigger the release job in the travis yml file? Is that simply creating a release in GitHub? Also, is the rpm artifact uploaded to the release page on GitHub by Travis CI?

wltjr:
Yes  the README.md can be removed from cpack. That is entirely up to you all if you want to include that in any dep or rpm. I just tend to included it as a packaging 101, add readme etc like autotools requires it. Cpack/Cmake does not require a readme to my knowledge. I could be wrong, removal will confirm either way.

To manually build dep or rpm, you do ```make package``` after running cmake, unless using ninja, then its ```ninja package```.  I need to confirm but I assume cpack is run and deb/rpm created when ```CMAKE_BUILD_TYPE=Release``` If that is the case, then making cpack optional would mean it only builds when you do ```make package``` regardless of ```CMAKE_BUILD_TYPE=Release```. I know you can create the package when the build type is other than release like debug by package target. To make a deb/rpm with debugging symbols. But seems its being run with the other builds, which I am pretty sure are not doing make package. If they are, then that is why.

In order to test out a release, need to do a tag on master branch. You could drop the conditional requirement for master branch. Then any tag should create one I think. The deploy section is always bound to a tag. But the conditional I added will only show that section and do that build when its going to upload. Otherwise it just builds with no upload, no point, thus conditional. Yes both deb and rpm will be uploaded to github by travis using your encrypted key under releases like this for [jem](https://github.com/Obsidian-StudiosInc/jem/releases) or [asspr](https://github.com/Obsidian-StudiosInc/asspr/releases). Then others like me can fetch if from there for use in travis or other. Which is my goal.

brarcher:
Go ahead and remove the README.md from cpack. If later someone really wants it the file can be added back. I'm not expecting the consumers of the archives will need it, though.

My main concern about when the archives would be built is to avoid the Windows builds from doing something odd. Those builds are currently working just fine, but now I wonder if the reason is they are building only in Debug mode.  I think it would be better to build cpack optionally instead of basing it on the build mode, just in case it could negatively impact those using CMake and Windows.

Is there a way to compute the tag instead of having the variable in the yml file? Currently the source of truth is at the [top of the configure.ac file](https://github.com/libcheck/check/blob/master/configure.ac#L8); the CMake build will [parse that file](https://github.com/libcheck/check/blob/master/CMakeLists.txt#L27) to determine the version. If there was a way to extract it in Travis-CI, it would be one less point of maintenance when cutting a release (and be less likely for me to mess it up).

The files being created by Travis-CI are labeled as x86_64 right now in the yml file. It is known that Travis-CI will always build on a x86_64 node?

wltjr:
Readme is gone, we will find out if its required. Pretty sure its optional. Some are picky about having such. I agree I see little need for it in the releases. It would already be read on the Github page or other if needed.

I can see about making the cpack optional or under some other cmake argument. It should be safe for windows, likely just ignored or something given cmake is cross platform. I think by default cpack is not optional, since its pre-build, its build system generation. I was mixing build with generation. Cmake just spits out files, doesn't build anything. Thus deb and rpm are only created with make/ninja package. But still I think we can make cpack section optional. I will experiment with such.

I was thinking about the version as it is one of my own annoyances. I have 2-3 places to update the version per release. I wanted to script i tall with git tag hooks. Though I wonder if a wildcard would suffice. We seem to be able to use variables, so why not just a * in place of the version. I think that would suffice. Or maybe a one liner that sets a variable after build in the build script section, to set the variable to what ever value of the generated files. That could be a backup to wildcard. I need to experiment with wildcard myself.

As for the x86_64 issue. For the time being Travis does not offer Linux other than x86_x64 to my knowledge. I added a conditional to not run that on other OS other than Linux. I think till Travis has something other than x86_64 for linux. That would be something to address then, but for now its not really of much concern. Not sure there is means easily to cross compile for other archs under Travis.

The filename of the deb/rpm version from cpack is completely configurable. This line
```cmake
set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}.${CMAKE_SYSTEM_PROCESSOR}")
```

I just go with the arch it was built on so others know what it will run on. Ideally it exists for more. But the main purpose of this would be so it run on Travis. We could look to do something for OS X so people could have a current libcheck for Travis OS X testing if you like. But that may already have access to the latest version.

Could drop the system processor from the filename entirely if you like. It is all up to your preference, your project. I just started with my personal defaults.

wltjr:
We could even tag the filenames with Travis or something to say its to be used under Travis. Though not sure any purpose to RPM at that point. It is all optional. I am not sure if you all would want to consider these official releases. Since not built for all platforms etc. Some projects use it for all official release for different Linux archs and different OS including Windows, like [Firebird](https://github.com/FirebirdSQL/firebird/releases) They have lots of release artifacts, downloadable pre-built binaries.

I am not sure if you all want to go that far. Eventually maybe something to consider if there is any need or justification. I just wanted to have a newer libcheck under Travis for my own projects. I could not find any third party ubuntu or debian packages of a newer libcheck. Thus coming to you all to make them directly. Though my luck after we do all thus, ubuntu and debian will update check package :face_with_head_bandage:

wltjr:
We may need to drop the entire deploy section from travis and handle the release upload differently. I see you are already doing releases now on Github, changelog and some tar.gz in addition to sources. This may conflict with that, or something, not sure. May need to just upload any deb/rpm along with the tar.gz.

brarcher:
My main development machine currently is a mac. If I'll need to start a VM to create the rpm and deb file, I'm more likely to forget or mess up.

Is the only option for Travis-CI to create the release?

wltjr:
It all depends on how you are doing the releases now. I just looked at the tar.gz, it seems to just be another copy of the sources. Not a pre-built binary as I was assuming. I guess something needed to make it officially a release.

Travis is not the only option. Travis is using Githubs api, so that is always an option. Though not to familiar with doing releases on github without Travis. There is a [manual release process](https://help.github.com/articles/creating-releases/). Where files can be dragged and dropped, but need to be built. Maybe that is what your doing now. Rather not have you have to fire up a VM for every release just for binaries.

I just do not want to mess up or clobber your existing release process. If the current release process is not automated via tag, then it is less of an issue. Just have to figure out the automated Changelog generation for release. Unless that can be added to the release after Travis uploads files.

brarcher:
The release process right now is documented [here](https://github.com/libcheck/check/blob/master/HACKING#L18). It uses the GitHub interface to manually tag a release. The tarball is built using autotools, and packages the configure script along with the source. I'll upload that tarball to the release page by dragging and dropping. That can be done on my mac, so it is not really an issue.

If Travis-CI kicks off the release process, what would that look like? What would I need to do to kick it off. I'm a little hesitant to give the start of the release process to a 3rd party provider, as if they have an issue or change some interface I'll need to figure that out to release again. Originally I was using CloudBees for automated testing, but eventually due to various changes on their and GItHub's side it become incompatible and I needed to migrate everything there to Travis-CI.

wltjr:
This would make your life much easier. All you do is create a tag in git and push it to github.
```
git tag 0.13.0
git push --tags
```
Then it builds under travis and uploads it all for you. You could stop doing your entire release process short of doing something for the changelog. It will make releases much quicker for you. Tag push and your done, short of changelog.

It will look the same as if you dragged and dropped the files for a release. The only difference is no changelog. Also no source tarball, because Github automatically creates those. Though CPack can generate one as well and Travis can upload it as well just like you do now.

I am not sure this integration would break anytime soon as many use it between Travis and Github. I would assume they coordinate on any API changes. It is a widely used feature by many projects.

Really 2 separate things,
 - Cmake making the stuff, deb, rpm, binary tar.gz, and source tar.gz
 - Travis building those binaries and uploading it to github

The 2nd part could be done regardless. You could be having Travis do the release for you now regardless of the cmake stuff.

I just find it a easy way to make deb and rpms for others when they do not exist. Also for my own needs, as I turn around and install the just built package via deb on travis. So its dual usage, self installation, and making binaries for others. Plus its a super fast way to do a release all around. Tag, push, and your done.

wltjr:
I read over the current release process and this would simplify that a bit.

- Step 5 could be removed, no need to build locally. Really if a previous CI run failed, then should not be doing a release. No need for a tarball to upload to Travis-CI. Github auto creates source tarballs on tag. Which are always present for releases. Travis_CI will be uploading the other generated deb and rpm. Thus no need to build anything locally.
- Step 7 would somewhat remain, but you would be [editting a release](https://help.github.com/articles/editing-and-deleting-releases/). Dropping all but lines [74-83](https://github.com/libcheck/check/blob/master/HACKING#L74), with no tarball to uploaded as mentioned on [line 85](https://github.com/libcheck/check/blob/master/HACKING#L85).

I can add an updated release procedure to this PR if you like.

brarcher:
I think you have sold me, or at least enough to try it out. Lets give it a shot.

> Also no source tarball, because Github automatically creates those.

The reason I was uploading a tarball is that the configure script would not exist in the GitHub's tarball. Would it be possible for the tarball with the configure script to be generated by Travis-CI as a part of the release, or is this mostly unnecessary in your opinion?

> Step 7 would somewhat remain, but you would be editting a release. Dropping all but lines 74-83, with no tarball to uploaded as mentioned on line 85.
> I can add an updated release procedure to this PR if you like.

Please do. (:

wltjr:
I think we can generate a tarball with the configure script. Your source tarball that includes that is generated via autotools. CMake uses either Make or Ninja as its generator and makes files for that given build system. But CMake does not generate a configure script. Likely need to use the current autotools stuff if you want a tarball with a configure script for build with make/autotools.

Seems you have 2 build systems, CMake and Autotools. When you do a release now, your source tarball contains both. Not sure if you want to continue to maintain 2 build systems or not. Its some what pointless as CMake can generate the necessary Make file like Autotools. Or you can keep having two that is your call. Seems an end user could use either CMake or Autotools.

I think for Autotools based projects, the source tarball does tend to contain a pre-generated configure. Though many tend to rely on running autoconf, automake, etc usually as part of autogen.sh script. Which does not exist in your repo or source tarball. Its just a wrapper script to invoke the auto tools and spit out configure script. Many projects still ship a pre generated configure script even if a autogen.sh exists as well in the repo/tarball. They leave it up to the end user to use the pre-generated one, or nuke and re-generate via autogen.sh/Autotools.

We can have Travis use the existing Autotools build system to generate a source tarball like you are now. Just generated on Travis instead of locally. Travis uploading to Github with deb and rpm. I can optionally add those files to CPack source tarball generation. Or we can and just include the source tarball from Autotools with the rpm and deb generated by CMake. A choice to be made there, not sure which is easiest. Maybe best to use existing plus CMake, That way if you decided to say remove Autotools build system, you do not need to modify CMake. Not sure it makes sense to have CMake generate a source tarball, given you still have to run autotools for the configure script.

I will update the release notes as part of the PR once we figure out the source tarball aspect.

brarcher:
> Seems you have 2 build systems, CMake and Autotools.

The project originally used Autotools and supported Unix-like OSs (Linux, MacOS, etc). Since then the project has expanded to officially support Windows platforms. It was found that Autotools would not work for this, so CMake was adopted to cover the Windows cases. There is stuff missing from the CMake builds that the Autotools builds have, and I've never invested the time to fully commit to CMake. Hence, the CMake builds officially support Windows and Autotools officially supports everything else. If someone were to volunteer to fully covert to CMake I'd help review the changes. (:

As for the source tarball, the only reason it would be provided as a download is because of the configure script. I agree it might be a bit odd for CMake to poke Autotools to generate the configure script. If there could be a separate step just for Autotools to generate the configure script, that might be fine.

wltjr:
Thanks for the back story. Makes sense as to why two build systems. If you are interested in dropping Autotools, I can see about finishing off what ever CMake is missing. Seems pretty close to complete since CMake will be used to generate the RPM and DEB.

For the time being I will just have Travis generate the source tarball using Autotools. I will just chain in Autotools before or after CMake. It will not be a step for someone to do, as it will be part of the automation via Tag. I will cover that in the updated release procedures. Then after we get the auto release via Tag taking place. You can decide on keeping or dropping Autotools. Then we just remove stuff vs add/modify. Like the commit I will make to chain in Autotools and generate source tarball. That commit could be reversed later if Autotools is dropped.

I will push some more commits here ASAP. Just been busy with some other projects. I will push 2 commits, 1 that has Autotools generate source tarball and upload via Travis. Then another that updates the release procedures.

wltjr:
The modifications to Travis.yml are pretty straight forward. Basically added old step 5, new step 6 to .travis.yml. But rather than running distcheck I am just running dist, since we just want the resulting tarball. Other tests would have already run before this. The tarball may even be generated already and we can drop all of this except the file.

I am not sure if you want to keep step 5/6. There really is no need to run make distcheck locally anymore. I left that in the updates, but we can drop that step. Travis will already be doing other builds. I added a step to Tag after a previous commit and push builds on Travis. Which if that fails that should be fixed. If it passes, just tag and push the tag to github. You cannot really push commits and tags together. That separation basically eliminates the need to run make distcheck locally. Pushing the commits becomes the testing step. With the tag step being the release step. Then from there its pretty much the same, but instead of drafting a new release, you are editing an existing tag you just pushed.

It ends up with 2 more steps. That is due to the version being in configure.ac and .travis.yml. A few ways to simplify that, like a wrapper script that sets version in both. Otherwise maybe some sed/awk in Travis to get the version from the configure.ac. That would take out that step. The other new step is the tagging step. That really cannot be eliminated. Though maybe already part of the procedures. I did not see an explicit tag. I assume that is done as part of a release. I guess it was being done as part of "Drafting a release" on Github, creating the tag then and there. This way the tag is created locally in git and pushed to Github. Which is what triggers the auto-release uploading stuff built by Travis, source tarball, rpm, and deb.

Anyway take a look, we can revise the release procedures a tad, drop the new step 6, old step 5. And drop the step to update travis.yml. I have a feeling you prefer to have 1 place to edit for versions. Seems CMake already gets that from Autotools. Something we would have to change if dropping Autotools. But could have Travis sed/awk the version from CMakeList.txt just the same as configure.ac.

I will go ahead and see about modifying travis.yml for auto version and dropping that step. Let me know on the other step, new 6, old 5. local building, make distcheck. Pretty sure that step can be dropped, but that is pending your feedback.

HACKING
```diff
@@ -40,36 +40,45 @@ CHECK_MICRO_VERSION=1
...
 Sun Aug 2, 2015: Released Check 0.10.1
   2015-08-02 19:21:14 +0000
   based on hash f399542eeceb97703bca496b68bb39044e8baa01
 
-4) Update index.html mentioning the release. Look for the following:
+5) Update index.html mentioning the release. Look for the following:
 <!-- Update this section during a release -->
 
-5) Attempt to build the release locally
+6) Attempt to build the release locally
```
brarcher: I agree that this step can be removed.

HACKING
```diff
@@ -40,36 +40,45 @@ CHECK_MICRO_VERSION=1
...
+7) Commit the changes to configure.ac, NEWS, and index.html to the Check 
+project's master branch, with the commit message:
 "Update for release"
 
-7) Log On to GitHub and navigate to:
+8) Push the commit to Github
 
-https://github.com/libcheck/check/releases
+9) If the build passes, tag the commit with the current release number 
+and push the tag to Github.
```
brarcher: Please put an example of creating the tag using git. The tag mentioned in the example from step 2 is 0.10.1, so use that in the example.

brarcher:
> If you are interested in dropping Autotools, I can see about finishing off what ever CMake is missing.

Sure, that would be great. Just let me know when you are up for it. (:

The points that come to mind are some configure --enable flags for some features, some configure checks for some platform support, and building the documentation.


> Just been busy with some other projects.

No worries. I appreciate the help.


> I am not sure if you want to keep step 5/6. There really is no need to run make distcheck locally anymore.

It used to be that there were no platforms supported but Linux, and running distcheck was the release check. Now, many tests are run on various platforms on Travis-CI and AppVeyor, so every merge request is "release ready" in that it must pass all tests before being merged in.

I agree, there is no need to run 'distcheck' locally. That step can be removed.


> I have a feeling you prefer to have 1 place to edit for versions. Seems CMake already gets that from Autotools. Something we would have to change if dropping Autotools. 

I'm not that worried about it. If Autotools is removed, something else could be the "source of truth". For example, maybe the header in the NEWS file is used as the current version, not sure. Or, CMake will contain the version, and anything that needs it will need to pull it from there.

> I will go ahead and see about modifying travis.yml for auto version and dropping that step.

I would appreciate it. (:

wltjr:
I still need to update travis so I can drop step 4 of updating .travis.yml. I also want to see if we even need to re-build the source tarball. I think a previous build job runs make distcheck or some target. If that is the case its likely generating the source tarball and we can use that. Semi-moot just double build will add a tad more time to build. Though if other tests are made optional on a tag, it would be better as is.

For my own projects I am likely going to not have test builds run on tags and just do deployment job. Since you have to push commits, and tags separately. I see it is push commits. If builds pass, then tag and push. At which point not much reason to re-build all over again and run test builds. It should be known to pass those, no changes, no commits. Just tagging a commit that was already pushed and passed all tests builds. Thus pushing a tag based on that good build. Again its moot, just more time in the release build that does the uploading. If you could push tags and commits, then its less of an issue. Still a build that fails, likely should never be tagged, much less uploaded for deployment/release.

PR #176, opened 2018-11-30
Fix typo

houstar:
Change 'iff' to 'if'


brarcher:
This case is not a typo. '[iff](https://en.wikipedia.org/wiki/If_and_only_if)' is short for 'if and only if'. That is used a few times in the file.

vtorri:
i think this PR can be closed

PR #189, opened 2019-05-31
Fix script to work from any path and to take argument

mikkoi:
Signed-off-by: Mikko Koivunalho <mikko.koivunalho@iki.fi>

### DIFF
--- a/tests/test_mem_leaks.sh
+++ b/tests/test_mem_leaks.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env sh
 
-UNIT_TEST=./check_mem_leaks
+SCRIPT_PATH="$(dirname "$(readlink -f "$0")")"
+if test -z "${1}"; then
+    UNIT_TEST="${SCRIPT_PATH}/check_mem_leaks"
+else
+    UNIT_TEST="${1}"
+fi
 VALGRIND_LOG_FILE=${UNIT_TEST}.valgrind
 LEAK_MESSAGE="are definitely lost"
 


brarcher:
Can you explain why this change is needed?

mikkoi:
Just to make operation easier for developer because the script has dependency on compiled programs in the same directory.
Developer can now execute the script from any directory.

```
~/check (master %=) $ tests/test_mem_leaks.sh
valgrind: ./check_mem_leaks: No such file or directory
No memory leaks found
~/check (master %=) $ git co improve-test_mem_leaks
Switched to branch 'improve-test_mem_leaks'
~/check (improve-test_mem_leaks %=) $ tests/test_mem_leaks.sh
==18845== Memcheck, a memory error detector
==18845== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==18845== Using Valgrind-3.10.0 and LibVEX; rerun with -h for copyright info
==18845== Command: /home/3/mikkoi/other/own_github/check_auto/tests/.libs/lt-check_mem_leaks
==18845==
^C
~/check (improve-test_mem_leaks %=) (130) $
```

makr:
I understand that I am a little late to the party. Please be aware that using ```readlink``` is a call for desaster, as this is not a POSIX utility and probably not available on comercial UNIX Systems. I need to verify this, however.

mikkoi:
Thank you for pointing this out. Do you know if there si any portable way of doing same the?

It is beside the point but we have to remember that this script calls on libtool which a GNU program (like readlink) and Valgrind, also not POSIX.

makr:
Ah! Should have looked into the whole script. Valgrind is indeed very Linux centric, so its probably OK to use GNU tools here. I noticed that you are doing a POSIX stunt to find a sh which will probably implement POSIX compatibility, but then use readlink. Which looked odd.
In order to have a working script you should check if libtool and valgrind are available before attempting something useful. Something like ```command -pv libtool``` can be helpful here.

mikkoi:
Something like this?
```
PATH_VALGRIND="$(command -pv valgrind)"
if test "0" -ne "$?"; then
    echo "Program 'valgrind' not installed."
    exit 127;
fi
PATH_LIBTOOL="$(command -pv libtool)"
if test "0" -ne "$?"; then
    echo "Program 'libtool' not installed."
    exit 127;
fi
```

makr:
Almost :-) ... ```"$(...)"``` may not work on Solaris, as /bin/sh is crippeled there. A better way - with a possibility to provide it as a parameter - would be something like:
```
VALGRIND="`command -pv ${VALGRIND_TOOL:-valgrind}`"
if [ -z "${VALGRIND:-}" ]; then
  printf "Program '${VALGRIND_TOOL:-valgrind}' is not installed.\n"
  exit -1
fi
```
This way one could set VALGRIND_TOOL and have it point to the path of the actual valgrind executable. If it is not set, it defaults to 'valgrind' which is then looked up along the PATH. If VALGRIND_TOOL is set, but not there, it would also result in an empty VALGRIND variable.
Same thing for libtool.

Issue #206, opened 2019-07-15
Is makeinfo required?


pierre-24:
Even though I do not want to build the documentation (because I'm using your library in a CI context, inside a docker container), I still need to install `texinfo`, otherwise I get 

```
WARNING: 'makeinfo' is missing on your system.
You should only need it if you modified a '.texi' file, or
any other file indirectly affecting the aspect of the manual.
You might want to install the Texinfo package:
http://www.gnu.org/software/texinfo/
The spurious makeinfo call might also be the consequence of
using a buggy 'make' (AIX, DU, IRIX), in which case you might
want to install GNU make:
http://www.gnu.org/software/make/
```

while running

```bash
autoreconf -i
./configure
make  # ! this fails
```

Not that I cannot install `texinfo`, but the TeX distribution is >1Gio, and I would like to keep the docker image as light as possible (I currently include `texiinfo`, but if I can get rid of it, that would be nice)

Thanks in advance :)

brarcher:
There is now a configure flag which controls if docs are built. See #212 for details. If that is not sufficient, please re-open.

PR #212, opened 2019-08-19
configure: optional build documentation

mikkoi:
Fix issue https://github.com/libcheck/check/issues/206

Applied an uncommitted patch by Branden Archer.
https://sourceforge.net/p/check/bugs/101/

Signed-off-by: Mikko Johannes Koivunalho <mikko.koivunalho@iki.fi>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -449,4 +449,11 @@ else
 fi
 echo "POSIX regular expressions ............ $result"
 
+if test "xtrue" = x"$enable_build_docs"; then
+    result="yes"
+else
+    result="no"
+fi
+echo "build docs ........................... $result"
+
 echo "=========================================="
--- a/configure.ac
+++ b/configure.ac
@@ -72,6 +72,15 @@ AC_HELP_STRING([--enable-timeout-tests],
   *)   AC_MSG_ERROR(bad value ${enableval} for --enable-timeout-tests) ;;
 esac], [enable_timeout_tests=true ])
 
+AC_ARG_ENABLE(build-docs,
+AC_HELP_STRING([--enable-build-docs],
+           [turn on building documentation @<:@default=yes@:>@]),
+[case "${enableval}" in
+  yes) enable_build_docs=true ;;
+  no)  enable_build_docs=false ;;
+  *)   AC_MSG_ERROR(bad value ${enableval} for --enable-build-docs) ;;
+esac], [enable_build_docs=true ])
+
 AM_CONDITIONAL(NO_TIMEOUT_TESTS, test x"$enable_timeout_tests" = "xfalse")
 
 AC_ARG_ENABLE(subunit,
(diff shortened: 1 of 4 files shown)


lib/snprintf.c
```diff
@@ -1097,6 +1097,9 @@ fmtflt(char *str, size_t *len, size_t size, LDOUBLE fvalue, int width,
 		infnan = (flags & PRINT_F_UP) ? "INF" : "inf";
 
 	if (infnan != NULL) {
+		int i = MAX_CONVERT_LENGTH - 1;
+		while (i >= 0) /* Initialize buffer, do not use memset (compatility) */
+			iconvert[i--] = '\0';
```
brarcher: Why not use memset?

Also, why not do this at the beginning at line 1084?

Should the same be done for fconvert, at least for consistency?

brarcher: Also (not a requirement, just a request), could you send the patch back to the original author as well?

http://www.jhweiss.de/software/snprintf.html

src/check_error.c
```diff
@@ -18,6 +18,17 @@
  * MA 02110-1301, USA.
  */
 
+/**
+ * We need the GNU C version to prevent using pragmas where they are
+ * not supported.
+ */
+#define GCC_VERSION (__GNUC__ * 10000 \
```
brarcher: Should this be in libcompat instead? Or, is one goal to reduce/remove that header eventually?

lib/libcompat.h
```diff
@@ -45,6 +45,11 @@
 #define CK_ATTRIBUTE_NORETURN
 #endif /* GCC 2.5 */
 
+/* Actually GCC 4.6.4 */
```
brarcher: Oh, you did move it to the header.

brarcher:
I see that the Travis-CI build stopped working on the optional docs change. Does there need to be another example build to check that the docs can still be built, and that the docs can be skipped?

lib/snprintf.c
mikkoi: Function memset() is not used anywhere in libcompat. So I figured maybe someone had already come to the conclusion that memset() is not available at every C installation or other compatibility issue. A little backwards logic here, I know. Just being overly careful.

However, memset() is used in libcheck in three places. If you say that it is okay to use memset(), I'm all for it.

I will move it to the beginning of the function, line 1084. A good idea, since the buffer is also used later in the function.

Same for fconvert? I will change that too.

lib/libcompat.h
mikkoi: I mean to improve the `GCC_VERSION_AT_LEAST(major, minor)` macro to `GCC_VERSION_AT_LEAST(major, minor, patch)` because the patch number is useful in some circumstances, such as here.

mikkoi:
Commit "configure: optional build documentation" does add a new feature, even if it is only in build.  @brarcher , can you add it to Travis?

brarcher:
The script which is run on Travis-CI is the travis.sh script. I'd suggest adding the following to the 

>   "${USE_CMAKE}" = 'NO' 

path around line 39:

> autoreconf -i || exit 1
> ./configure ${EXTRA_ARGS} --disable-build-docs|| exit 1
>  make || exit 1
> 
>    if [ -f doc/version.texi ]; then
>       echo "Documentation was generated, but was disabled";
>       exit 1;
>    fi

The rest of the test can proceed as normal.

mikkoi:
Missing the issue https://github.com/libcheck/check/issues/217 .
Should we postpone?

brarcher:
If you would like to postpone merging this until both CMake and Autotools both have a configurable option to build the docs, that is fine. I'll leave it to your discretion. Just let me know when it is ready to merge.

mikkoi:
The documentation itself needs to be revised. Part of the example needs to be rewritten so show how the new CMake import works. I have improved the file FindCheck.cmake to follow new CMake best practices. It is used to find Check installation from CMake build when the build does not support CMake style export/import or the using build wants to use an earlier (0.12.0 or earlier) Autotools built installation.
I think we should merge this branch now and keep the issue #217 open but close issue #206 .

configure.ac
```diff
@@ -171,12 +180,19 @@ esac
...
 AC_CHECK_PROGS(GENHTML, genhtml, false)
-AC_CHECK_PROGS(TEX, tex, false)
-if test "$TEX" = "false"; then
-    # Make it [somewhat] clear to maintainers that tex is missing. Not an error
-    # though because 'make install' (which users need) does not build the docs
-    # anyway.
-    AC_MSG_WARN(tex not installed: cannot rebuild HTML documentation.)
+
+if test "xtrue" = x"$enable_build_docs"; then
+    AC_CHECK_PROGS(TEX, tex, false)
```
nmeum: Why does this check for tex instead of makeinfo? The issue you referenced also mentions makeinfo, not tex.

brarcher: Thanks for pointing this out, and thanks for your fix in https://github.com/libcheck/check/pull/231

PR #213, opened 2019-08-19
Fix some issues that came up with Travis updating to Clang 7

mikkoi:


lib/snprintf.c
```diff
@@ -1097,6 +1097,9 @@ fmtflt(char *str, size_t *len, size_t size, LDOUBLE fvalue, int width,
 		infnan = (flags & PRINT_F_UP) ? "INF" : "inf";
 
 	if (infnan != NULL) {
+        int i = MAX_CONVERT_LENGTH - 1;
+        while (i >= 0)
+            iconvert[i--] = '\0';
```
brarcher: Why not simply memset()?

brarcher: Oh, I now see your response in an email from GitHub. I'd change to using memset(), then I'll merge in these changes.

PR #225, opened 2019-10-20
Fix c standard to c99

mikkoi:
Fix required C standard from C90 to C99
    
    We are using C99 features, such as anonymous variadic macros and
    mixed declarations and code.
    
    Besides, we are using POSIX extensions (fork, etc.), so there is
    (probably?) no point in forcing the use of the old ANSI C89 / ISO C90.


brarcher:
Curious, I took a look at the autotools setup to see what it was doing. There was no reference there about forcing C89. Instead, there was a reference for configuring C99 for one particular compiler (as I guess its default was C89). C99 sounds good to me.

PR #229, opened 2019-10-20
Add compiler warnings for GNU and Clang

mikkoi:
When building CMake Debug.


brarcher:
Does you know if the [Travis-CI](https://github.com/libcheck/check/blob/master/travis.sh) or [AppVeyor](https://github.com/libcheck/check/blob/master/appveyor.yml) builds are Release or Debug? That is, will these flags be used in the default build for Check, and is there any test coverage on Travis-CI or AppVeyor for using the flags if it is not the default?

mikkoi:
Travis-CI builds are all 'Debug' unless specifically defined as 'Release' (currently not). The warnings are visible in Travis logs. They are only warnings, so the builds will not fail.

The purpose is the same as with running Cppcheck during build: emit all the warnings we can to inform the developer.

PR #232, opened 2019-10-25
Rename cmake project from 'check' to 'Check'

mikkoi:
Rename CMake project from 'check' to 'Check'
    
    Check should be written with a capital letter.
    In a CMake script this makes a difference as
    `find_package(Check)` should work, not `find_package(check)`.
    
    This change does not alter how Check is installed.
    Maintaining compatibility with GNU Autotools installation
    is mandatory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,13 +485,12 @@ endif (BUILD_TESTING)
 
 ###############################################################################
 # Export project, prepare a config and config-version files
-set(LIB_INSTALL_DIR lib CACHE FILEPATH "lib INSTALL DIR")
 string(TOLOWER ${PROJECT_NAME} EXPORT_NAME)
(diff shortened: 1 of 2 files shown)


brarcher:
If I understand the change, it will not affect CMake users' projects if they upgrade Check to this change, is that right? For example, could updating Check result in it no longer being found for a CMake project in Windows due to the name change?

mikkoi:
This change will not alter the behaviour how CMake projects are currently discovering Check. Since we have not supported CMake's own discovery methods before, including Cmake Package Registry, current users are using the Find*.cmake file method, as described in our documentation (:doc/example/cmake/FindCheck.cmake). This Find file uses the `pkg-config` program to find Check, and `pkg-config` only relies on library file names.

Issue #238, opened 2020-01-07
Getting things to work with CMake's FetchContent_*


unindented:
In my C++ projects, I'm used to including libraries like `googletest` and `catch2` using `FetchContent_*` functionality in CMake:

```
FetchContent_Declare(
  catch2
  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
  GIT_TAG        v2.10.2
)
FetchContent_MakeAvailable(catch2)

target_link_libraries(mytests PRIVATE myapp Catch2::Catch2)
```

I haven't been able to get a similar setup working with `check`. If I do something like this:

```
FetchContent_Declare(
  check
  GIT_REPOSITORY https://github.com/libcheck/check.git
  GIT_TAG        0.13.0
)
FetchContent_MakeAvailable(check)

target_link_libraries(mytests PRIVATE myapp Check::check)
```

I get this error:

```
CMake Error at tests/CMakeLists.txt:10 (add_executable):
  Target "loxtest" links to target "Check::check" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

What am I doing wrong?

mikkoi:
Check tries to be compatible with CMake but importing the project this way fails. Checked it myself.
You need to build check by itself. No need to install.
Then use find_package() to find it.

When building check (cmake <dir> && make), CMake registers the build in Package Registry (~/.cmake/package/). find_package() uses this.

Using ExternalProject_Add() in CMake might work (have not tested) because it builds the project independently from your main project and "installs" it in the build tree.

unindented:
What would it take for Check to be compatible with `FetchContent`? It's super convenient for the consumer.

mikkoi:
It certainly looks convenient. I'll take a look at it. For the time being Check needs to retain compatibility with GNU Autotools and that poses some challenges.

mikkoi:
I believe I added the necessary parts. Please try with branch mikkoi:fix-for-cmake-fetchcontent, pull request https://github.com/libcheck/check/pull/239 if it works for you aswell.

unindented:
@mikkoi thank you!

When I point `FetchContent` to your branch, CMake seems to complain about a submodule in the `tests/` folder (it says it's not present in the `.gitmodules` file). Do you know what that's about.

mikkoi:
[...]

PR #239, opened 2020-01-10
Fix for cmake fetchcontent

mikkoi:
Fix for issue https://github.com/libcheck/check/issues/238 and some cleaning.
Set policy CMP0090 to avoid unintentionally dirtying user's locale CMake Package Registry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 # Copyright (C) 2016 Branden Archer <b.m.archer4@gmail.com>
 # Copyright (C) 2016 Joshua D. Boyd <jdboyd@jdboyd.net>
+# Copyright (C) 2020 Mikko Koivunalho <mikko.koivunalho@iki.fi>
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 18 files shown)


brarcher:
I've not heard of FetchContent before. It sounds rather useful. Thanks for looking into this and enabling support for it?

Is there anything that can be done to add testing to ensure that it continues working in the future?

mikkoi:
I created a test in Travis. It builds an example project which uses Check in two different ways. Build script in 'travis.sh`.

unindented:
@mikkoi what's the purpose of this submodule in the diff of this PR?

Screen Shot 2020-01-13 at 1 58 58 PM

mikkoi:
tests/cmake_project_usage_test is used to test the solution. It is only done in Travis CI. That's why it isn't actually part of CMake build, just a separate CMake project.

mikkoi:
Marvelous! Travis is running CMake 3.12. FetchContent is supported but FetchContent_MakeAvailable() is not available until CMake 3.14. ... Time for some backwards compatibility spaghetti! :-)

unindented:
@mikkoi haha yup, this is what I have to do in my projects:

```
include(FetchContent)
if(${CMAKE_VERSION} VERSION_LESS 3.14)
  include(cmake/FetchContent_MakeAvailable.cmake)
endif()
```

Where `cmake/FetchContent_MakeAvailable.cmake` contains something like this:

```
macro(FetchContent_MakeAvailable NAME)
  FetchContent_GetProperties(${NAME})
  if(NOT ${NAME}_POPULATED)
    FetchContent_Populate(${NAME})
    add_subdirectory(${${NAME}_SOURCE_DIR} ${${NAME}_BINARY_DIR})
  endif()
endmacro()
```

mikkoi:
Thank you. That is a good solution. And very Cmake like, I think.

unindented:
I've pointed my project to your branch, and it works correctly!

PR #243, opened 2020-01-20
Add Repology badge to README.md

mikkoi:
Check is well represented. Let's show it.


brarcher:
Interesting. Never heard of the repology service before.

Issue #242, opened 2020-01-19
Getting things to work with CMake's FetchContent - Disable export (installation)


mikkoi:
Continuation to issue https://github.com/libcheck/check/issues/238

CMake FetchContent works but produces a surprising side effect: when installing the main project, sub project Check also gets installed in the same place, env `CMAKE_INSTALL_PREFIX`. This should be disabled by default. However, I am thinking if there are situations when this behavior is wanted, perhaps if the main project is using Check as a shared library. (?)

PR #244, opened 2020-01-23
Fix for cmake fetchcontent #2

mikkoi:
Fix for https://github.com/libcheck/check/issues/242


CMakeLists.txt
```diff
@@ -377,12 +377,14 @@ endif (HAVE_SUBUNIT)
 
 ###############################################################################
 # Generate "config.h" from "cmake/config.h.in"
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/config.h.in
-  ${CMAKE_CURRENT_BINARY_DIR}/config.h)
-  # Param @ONLY not used on purpose!
-include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
-add_definitions(-DHAVE_CONFIG_H)
-set(CONFIG_HEADER ${CMAKE_CURRENT_BINARY_DIR}/config.h)
+# if (THIS_IS_MAIN_PROJECT)
```
brarcher: Why is this one commented out?

CMakeLists.txt
```diff
@@ -397,64 +399,68 @@ set(CONFIG_HEADER ${CMAKE_CURRENT_BINARY_DIR}/config.h)
 #
 # When converting to CMake we also want to abandon the m4 macros.
 #
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/check_stdint.h.in
-  ${CMAKE_CURRENT_BINARY_DIR}/check_stdint.h @ONLY)
-install(
-  FILES ${CMAKE_CURRENT_BINARY_DIR}/check_stdint.h
-  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-)
+  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/check_stdint.h.in
```
brarcher: Why change the indentation on this one?

src/CMakeLists.txt
```diff
@@ -181,11 +181,15 @@ target_include_directories(checkShared
...
 )
 
-install(TARGETS check checkShared
-  EXPORT check-targets
-  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-  PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-)
+if (THIS_IS_MAIN_PROJECT)
```
brarcher: If the Check is not the main project then its library is not installed. Does this mean that the main project is not expected to install anything that depends on Check. For example, the main project is not installing its unit tests. (This makes sense, just checking if that is the rationale).

CMakeLists.txt
mikkoi: Fixed.

CMakeLists.txt
mikkoi: Fixed. Thanks for noticing.

src/CMakeLists.txt
mikkoi: Yes. In this matter I have followed the solutions in Catch2 and Googletest.

mikkoi:
I changed the variable name from THIS_IS_MAIN_PROJECT to THIS_IS_SUBPROJECT. It should highlight the fact that subproject is the exception, not the norm.

PR #249, opened 2020-02-13
Fix small issues 5

mikkoi:
Fix some warnings


lib/libcompat.h
```diff
@@ -48,8 +48,10 @@
 
 #if GCC_VERSION_AT_LEAST(2,95,3)
 #define CK_ATTRIBUTE_UNUSED __attribute__ ((unused))
+#define CK_ATTRIBUTE_FORMAT(a, b, c) __attribute__ ((format (a, b, c)))
```
brarcher: What this attribute introduced in GCC 2.95.3? I could not find a reference for that. Do you know when it was introduced? 

Or, should there be a configure check for it? Not sure what it would be for CMake, but this may work for autotools:

https://www.gnu.org/software/autoconf-archive/ax_gcc_func_attribute.html

tests/check_check_master.c
```diff
@@ -943,11 +943,13 @@ void record_failure_line_num(int linenum)
...
    linenum += 1;
 
-  to_write = snprintf(string, sizeof(string), "%d\n", linenum);
-  if(to_write == 0)
+  chars_printed = snprintf(string, sizeof(string), "%d\n", linenum);
+  if(chars_printed <= 0 || (size_t) chars_printed >= sizeof(string))
   {
     fprintf(stderr, "%s:%d: Error in call to snprintf:", __FILE__, __LINE__);
     exit(1);
+  } else {
```
brarcher: small point, the {} in the file follow this convention:

if() 
{
// stuff
}

Probably this else is not needed, though, as the if() bails out. Probably the to_write assignment can be outside of the else block.

brarcher:
May you add a note to the NEWS file for these changes, if they should be mentioned in the release notes? Probably the __attribute__ format change is worth mentioning as it is part of the API.

lib/libcompat.h
mikkoi: GNU 2.95.3 is the oldest version "supported" by GNU. I don't know what their official view is but 2.95.3 is the oldest version with documentation on GCC's pages, and __attribute__ ((format)) is supported by 2.95.3. All docs: https://gcc.gnu.org/onlinedocs/
Regarding __attribute__ ((format)): https://gcc.gnu.org/onlinedocs/gcc-2.95.3/gcc_4.html#SEC84


I would not go for a configure check because this is compiler specific thing and can be checked with compiler's versions.

tests/check_check_master.c
mikkoi: I moved it outside the else block. It's clearer that way. Thank you.

lib/libcompat.h
mikkoi: Actually, thinking about it, it would be good if there was a way to check the compiler's support for *attribute ((format))* in the build system, both Autotools and CMake. After all, Clang probably supports all the GNU C originated function attributes.

I think this should go to TODO...

brarcher: Really I'm concerned about compilers other than gcc and clang. I'd suspect that the way it is now the other compilers will not reach the format attribute as they would not claim to be GCC. 

A TODO is fine

mikkoi: Yes. Other compilers would get nothing instead of "__attribute__ ((format))" when macro CK_ATTRIBUTE_FORMAT is opened.

PR #250, opened 2020-03-01
Fix src/check_pack.c

mikkoi:
Some refactoring done in `src/check_pack.c`. The interface `src/check_pack.h` is unchanged. Because of this, I have added several casts. I have tried to work around the problems of signed/unsigned arithmetic.

### DIFF
--- a/src/check_pack.c
+++ b/src/check_pack.c
@@ -95,8 +95,8 @@ static void check_type(int type, const char *file, int line);
 static enum ck_msg_type upack_type(char **buf);
 static void pack_type(char **buf, enum ck_msg_type type);
 
-static int read_buf(FILE * fdes, int size, char *buf);
-static int get_result(char *buf, RcvMsg * rmsg);
+static size_t read_buf(FILE * fdes, size_t size, char *buf);
+static size_t get_result(char *buf, RcvMsg * rmsg);
 static void rcvmsg_update_ctx(RcvMsg * rmsg, enum ck_result_ctx ctx);
 static void rcvmsg_update_loc(RcvMsg * rmsg, const char *file, int line);
 static RcvMsg *rcvmsg_create(void);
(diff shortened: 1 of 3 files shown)


src/check_pack.c
```diff
@@ -70,12 +72,13 @@ static size_t get_max_msg_size(void)
...
 
-/* typedef an unsigned int that has at least 4 bytes */
+/* typedef an unsigned int that has exactly 4 bytes, as required by pack_int() */
 typedef uint32_t ck_uint32;
+#define CK_UINT32_MAX UINT32_MAX
 
 
-static void pack_int(char **buf, int val);
-static int upack_int(char **buf);
+static void pack_int(char **buf, ck_uint32 val);
```
brarcher: I'm wondering what the value of having the ck_uint32 typedef is, rather than just using uint32_t directly. I'm not sure it buys anything. Same for UINT32_MAX vs CK_UINT32_MAX. What do you think?

src/check_pack.c
```diff
@@ -196,7 +202,11 @@ static char *upack_str(char **buf)
     char *val;
     int strsz;
 
-    strsz = upack_int(buf);
+    ck_uint32 str_sz = upack_int(buf);
+    if(str_sz > INT_MAX)
```
brarcher: The check for the pack and unpack are not the same. A check for UINT32_MAX is done for the pack, but a check for INT_MAX is done for the unpack. Should the check be the same, and use the smaller of UINT32_MAX or INT_MAX?

brarcher: I see the point looking at some of the other commits that do the same thing. It is not about catching an issue when the value is first packed, but doing a sanity check before the value is used. That should be fine. It catches an issue before the data is used improperly.

src/check_pack.c
mikkoi: Well, `ck_uint32` was there when I came. :grin: 
I added `CK_UINT32_MAX` for the sake of consistence.
`ck_uint32` is only used in `check_pack.c` and only internally, not in the header. It would not be a big thing to refactor it away.

src/check_pack.c
mikkoi: Yes. It checks that the value is not too big. In fact, I am unhappy about writing a ck_uint32 to an int, because int's size could be smaller (and system dependent). In fact, if int is 32 bit, it is certainly too small because of the unsigned to signed conversion.
But DurationMsg is part of the API (the public header). If that were changed, it would remove the need for many checks.

brarcher:
Note to self: Travis-CI deprecated how it reports status back on builds. A build was triggered and passed: [link](https://travis-ci.org/libcheck/check/builds/659989763).

PR #255, opened 2020-04-20
Add reference to check-gen tool

oxr463:


brarcher:
I appreciate your interest in the Check library and in creating a downstream utility for creating unit tests. 

To my knowledge the only place that Check's documentation mentions other projects is where it lists other unit tests platforms, as a background to unit testing in general:

https://libcheck.github.io/check/doc/check_html/check_2.html#Unit-Testing-in-C

I'd like to avoid maintaining a list of downstream projects which use Check. Mentioning these projects could be considered an endorsement. As I am not familiar with check-gen and do not use it myself I'd like to avoid linking to it.

Again, thanks for your interest in the project, and for helping Check be more useful to more users.

PR #257, opened 2020-05-22
[windows] Implement pthread_mutex for Windows

walac:
When compiling for Windows, pthread libraries may not be available. For
Win32 applications in which an external implementation of pthread is not available, we build libcompat with the pthread_mutex.c module, which provides a simple implementation of pthread_mutex for the Win32 API.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,8 @@
 #
 cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
 
+include(CheckCCompilerFlag)
+
 # Detect if Check is being used in another build as a subproject
 # probably with command FetchContent*().
 set(THIS_IS_SUBPROJECT FALSE)
@@ -112,7 +114,7 @@ if(WIN32)
   if(MSVC60)
   set(WINVER 0x0400)
   else()
-  set(WINVER 0x0500)
+  set(WINVER 0x0600)
   endif()
   set(_WIN32_WINNT ${WINVER})
 endif(WIN32)
@@ -162,6 +164,10 @@ ck_check_include_file("strings.h" HAVE_STRINGS_H)
 ck_check_include_file("sys/time.h" HAVE_SYS_TIME_H)
 ck_check_include_file("time.h" HAVE_TIME_H)
 ck_check_include_file("unistd.h" HAVE_UNISTD_H)
+ck_check_include_file("pthread.h" HAVE_PTHREAD)
+
+# check if we have windows.h on native windows environments
+ck_check_include_file("windows.h" HAVE_WINDOWS_H)
 
 ###############################################################################
 # Check functions
@@ -180,6 +186,10 @@ check_function_exists(strsignal HAVE_DECL_STRSIGNAL)
 check_function_exists(_getpid HAVE__GETPID)
 check_function_exists(_strdup HAVE__STRDUP)
 check_function_exists(alarm HAVE_DECL_ALARM)
+if (HAVE_WINDOWS_H)
+  check_function_exists(InitOnceBeginInitialize HAVE_INIT_ONCE_BEGIN_INITIALIZE)
+  check_function_exists(InitOnceComplete HAVE_INIT_ONCE_COMPLETE)
+endif()
 if (HAVE_REGEX_H)
   check_function_exists(regcomp HAVE_REGCOMP)
   check_function_exists(regexec HAVE_REGEXEC)
@@ -232,6 +242,14 @@ if(HAVE_REGEX_H AND HAVE_REGCOMP AND HAVE_REGEXEC)
   set(HAVE_REGEX 1)
   add_definitions(-DENABLE_REGEX=1)
   set(ENABLE_REGEX 1)
+endif()
+
+if (HAVE_PTHREAD)
+  check_c_compiler_flag("-pthread" HAVE_PTHREADS_FLAG)
+  if (HAVE_PTHREADS_FLAG)
+    add_definitions("-pthread")
+    add_link_options("-pthread")
+  endif()
 endif()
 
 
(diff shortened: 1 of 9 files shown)


brarcher:
Can you give some background on what versions or variations of Windows does not have pthread? There are tests run on AppVeyor, so at least some version of Windows has it. Though, I'm not familiar with what version of Windows is actually used there. At least the compiler versions we test against expose pthreads.

If Check were to support Windows not having pthread we would need a way to test it. That is, some confidence the lack of pthread can be properly detected and when build to use critical sections the result works correctly.

src/check_pack.c
```diff
@@ -33,13 +34,33 @@
...
 #ifndef HAVE_PTHREAD
-#define pthread_mutex_lock(arg)
-#define pthread_mutex_unlock(arg)
 #define pthread_cleanup_push(f, a) {
 #define pthread_cleanup_pop(e) }
 #endif
 
+#ifndef HAVE_PTHREAD
+#if defined(HAVE_WIN32_INIT_ONCE)
+#define pthread_mutex_lock(arg) EnterCriticalSection(arg)
```
brarcher, line 56: Replacing the definition of the pthread methods seems a little magical to me. Reading the code later on it will not be obvious that a critical section is being used instead of pthreads.

I think it would be better to have a level of abstraction here about mutual exclusion. In Check's source there are several re-implementations of some APIs that may be missing or may use one of multiple implementations:

https://github.com/libcheck/check/tree/master/lib

The high resolution timer support is one example:

https://github.com/libcheck/check/blob/master/lib/timer_settime.c

There, if the high resolution calls are not available replacements are used.

Perhaps there can be some functions created that abstract away the details for pthread and critical sections.

walac:
> Can you give some background on what versions or variations of Windows does not have pthread? 

afaict, Windows doesn't natively support `pthreads`. So, unless the client uses a third-party implementation of pthreads for Windows, the code won't be protected by a mutex. And that's is exactly the situation we are facing, we are porting the [EFL](https://www.enlightenment.org/about-efl.md) to [Windows](https://github.com/expertisesolutions/efl), using `libcheck` for tests. For thread tests, an assertion is raised inside the C runtime when `libcheck` calls`fwrite`, due to the `FILE *`  being called from multiple threads without a protecting mutex.

src/check_pack.c
walac: Makes sense, I can implement a basic pthread_mutex inside lib on top of Win32.

brarcher:
Thanks for that explanation. It did not occur to me that Windows would not have pthreads natively, though that does make sense. So all this time the protections have been no-ops.... A bit of a miss -- thanks for pointing it out and volunteering to fix it!

If that is the case, Check's existing automated tests on AppVeyor should provide coverage. Disregard that part of my earlier comments.

Given that not having either pthreads caused an issue that Check was unable to detect in its unit tests, maybe it should be a configure-time failure if pthreads (and now critical sections) is unavailable. Or, maybe at least a warning of some sort.

I see that I did not finish the [migration](https://github.com/libcheck/check/pull/251) from Travis-CI to Github Actions, so the Linux and OSX automated tests did not run for your PR. I'll finish that up now. In a bit I'll let you know, so you can update your branch to pickup testing for Linux and OSX.

brarcher:
I've enabled the Linux and OSX tests now on Github. Update your branch to pick them up.

CMakeLists.txt
```diff
@@ -112,7 +114,7 @@ if(WIN32)
   if(MSVC60)
   set(WINVER 0x0400)
   else()
-  set(WINVER 0x0500)
+  set(WINVER 0x0600)
```
brarcher, line 117: I'm not familiar with Windows' versioning. Can you explain this change?

src/check_pack.c
```diff
@@ -33,13 +34,20 @@
 #include "check_impl.h"
 #include "check_pack.h"
 
+#ifdef HAVE_WINDOWS_H
+#include <windows.h>
```
brarcher, line 38: What does windows.h provide for this file?

lib/pthread_mutex.c
```diff
@@ -0,0 +1,70 @@
...
+    int ret = 0;
+
+    if (!InitOnceBeginInitialize(&mutex->init, 0, &pending, NULL))
+        return -1;
+
+    if (pending)
+    {
+        mutex->mutex = CreateMutexW(NULL, FALSE, NULL);
+        if (!mutex->mutex)
+            ret = -1;
```
brarcher, line 37: The pattern elsewhere in the code is to use {} around single line if() statements
```
if (/* condition */)
{
  // stuff
}
```

brarcher: oh, I guess that is not true, later in this review ppack.c is doing just this. Nevermind then.

brarcher:
I appreciate your adding support for both Autotools and CMake. (:

CMakeLists.txt
walac: The [One-Time Initialization](https://docs.microsoft.com/en-us/windows/win32/sync/one-time-initialization) functions were introduced in Windows Vista. The `WINVER` value for Windows Vista is [0x0600](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019).

src/check_pack.c
walac: Ah, this was an artifact from the previous patch. Removed, thanks.

brarcher:
Thanks for implementing this!

walac:
@brarcher Is it possible we get a new release of the library? I would like to update the [meson wrapdb](https://wrapdb.mesonbuild.com/check).

brarcher:
Sure. I'll create a release today.

brarcher:
[Check 0.15.0](https://github.com/libcheck/check/releases/tag/0.15.0) has been released.

walac:
> Check 0.15.0 has been released.

Thanks a lot \o/

PR #263, opened 2020-06-12
Add ck_skip() and ck_skip_msg(...) macros [WIP] [DISCUSS]

### LINKED ISSUE #178: Feature request: skipping a test from within a test
I'd like something like `ck_skip()` that skips the currently running test. Usage would be:

```
START_TEST(foo)
{
    if (!something)
       ck_skip()

   ...
   ck_assert(...)
}
END_TEST
```
Obviously a simple `return`  has the same effect, but a `ck_skip()` macro could log/process the skipped test better, e.g. warn if a test was called multiple times but skipped every time.

I admit this is a niche use-case. I'm using check for [libinput](https://gitlab.freedesktop.org/libinput/libinput) and the test suite has a slightly peculiar design. We have test uinput devices and the test suite adds each test for each matching test device. So e.g. a touchpad tapping test is called for all touchpad test devices.

But some of those devices have special run-time properties that we don't necessarily know at test setup time. So what I need is the ability to skip from within *and* some way to access that information afterwards. I want to check if a test was skipped for **every** test device, because then there's some other bug somewhere. Example again: if I test for right buttons on touchpads, I want to skip this on touchpads that don't have right buttons. But if the test always skips, I either have wrong test devices or something else is buggy.

Not 100% how this would look like from check's perspective though, but let's see if this is something wanted/possible at all. If not, that's fine, I agree that this is a bit of a special case :)

jbboehr:
* `ck_skip()` aborts the test and marks it as skipped
* `ck_skip_msg(...)` does the same with a custom printf message
* `_ck_skip()` is exposed as a symbol

These are all analogues of `ck_abort()`, `ck_abort_msg()`, and `_ck_assert_failed()`

Since they basically work like `ck_abort()`, I decided to use a flag `wasskipped` in `FailMsg` instead of adding a new message type.

As it is, I'm not including skipped tests in the success percentage, but they are included in total tests run. I considered adding `srunner_ntests_skipped()` and `srunner_skipped()` but decided to hold off since they are included in `srunner_results()`

In `tr_type_str`, I assigned skip to `K` because the default return value was `S` already. I'm not sure if there's a more appropriate value here.

I'm personally using check with the autotools tap driver, and it seems to be working alright so far.

Let me know if you have any thoughts and, of course, feel free to request changes.

Closes #178

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -362,7 +362,10 @@ void _mark_point(const char *file, int line)
     send_loc_info(file, line);
 }
 
-void _ck_assert_failed(const char *file, int line, const char *expr, ...)
+#if HAVE_FORK
+CK_ATTRIBUTE_NORETURN
+#endif
+static void _assert_failed(const char *file, int line, int wasskipped, const char *expr, va_list apo)
 {
     const char *msg;
     va_list ap;
@@ -371,7 +374,11 @@ void _ck_assert_failed(const char *file, int line, const char *expr, ...)
 
     send_loc_info(file, line);
 
-    va_start(ap, expr);
+#if defined(va_copy)
+    va_copy(ap, apo);
+#else
+    ap = apo;
+#endif
     msg = (const char *)va_arg(ap, char *);
 
     /*
@@ -389,7 +396,7 @@ void _ck_assert_failed(const char *file, int line, const char *expr, ...)
     }
 
     va_end(ap);
-    send_failure_info(to_send);
+    send_failure_info(to_send, wasskipped);
     if(cur_fork_status() == CK_FORK)
     {
 #if defined(HAVE_FORK) && HAVE_FORK==1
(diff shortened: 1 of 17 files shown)


thom311:
I think a skip should not abort the test, unlike an assertion failure. Otherwise, if you run without fork-mode, an earlier skip will prevent all followup tests from running. With assertion failures that is fine, because you better debug the first failure you encounter. For skip, that seems undesirable.

Compare to glib's [`g_test_skip()`](https://gitlab.gnome.org/GNOME/glib/-/blob/b9490a499a004618c883f180b1081a166ff1a86b/glib/gtestutils.c#L2582), which just remembers the verdict of the test, and evaluates the test result at the end.

Yes, it means, usually the user would follow up that with an additional return, like
```
if (!has_dependency()) {
     ck_skip("Lacks dependency");
     return;
}
```

I think this would be a good feature.

Issue #270, opened 2020-06-22
windows with msys : warnings when printf functions


vtorri:
Here are the warnings : 

```
check_error.c: In function 'emalloc':
check_error.c:64:29: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
   64 |         eprintf("malloc of %u bytes failed:", __FILE__, __LINE__ - 2, n);
      |                            ~^                                         ~
      |                             |                                         |
      |                             unsigned int                              size_t {aka long long unsigned int}
      |                            %llu
check_error.c: In function 'erealloc':
check_error.c:74:30: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
   74 |         eprintf("realloc of %u bytes failed:", __FILE__, __LINE__ - 2, n);
      |                             ~^                                         ~
      |                              |                                         |
      |                              unsigned int                              size_t {aka long long unsigned int}
      |                             %llu
check_pack.c: In function 'pack':
check_pack.c:139:33: warning: format '%d' expects argument of type 'int', but argument 4 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
  139 |         eprintf("Value of len (%d) too big, max allowed %u\n",
      |                                ~^
      |                                 |
      |                                 int
      |                                %lld
  140 |                 __FILE__, __LINE__ - 3, len, INT_MAX);
      |                                         ~~~
      |                                         |
      |                                         size_t {aka long long unsigned int}
check_pack.c: In function 'upack':
check_pack.c:162:35: warning: unknown conversion type character ')' in format [-Wformat=]
  162 |         eprintf("Value of diff (%t) too big, max allowed %u\n",
      |                                   ^
check_pack.c:162:59: warning: format '%u' expects argument of type 'unsigned int', but argument 4 has type 'ptrdiff_t' {aka 'long long int'} [-Wformat=]
  162 |         eprintf("Value of diff (%t) too big, max allowed %u\n",
      |                                                          ~^
      |                                                           |
      |                                                           unsigned int
      |                                                          %llu
  163 |                 __FILE__, __LINE__ - 3, diff, INT_MAX);
      |                                         ~~~~               
      |                                         |
      |                                         ptrdiff_t {aka long long int}
check_pack.c:162:17: warning: too many arguments for format [-Wformat-extra-args]
[...]

PR #271, opened 2020-06-22
Fix format specifiers that do not match the argument types

jamesjer:
In check.h.in, %x is not safe to use with pointers, since it specifies an int-sized value but a pointer is passed in.  On some platforms, `int` is 32 bits and `void *` is 64 bits.

In check_pack.c, note that `t` is a length modifier, not a conversion specifier.  It must be followed by either `d` or `u`; `d` seems appropriate in this case.

The other changes are fairly obvious mismatches between the argument type and conversion specifier.

### DIFF
--- a/src/check.h.in
+++ b/src/check.h.in
@@ -1724,7 +1724,7 @@ do { \
 #define _ck_assert_ptr(X, OP, Y) do { \
   const void* _ck_x = (X); \
   const void* _ck_y = (Y); \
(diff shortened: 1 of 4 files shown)


jamesjer:
This commit fixes https://github.com/libcheck/check/issues/270

brarcher:
Thanks for the fixes!

vtorri:
it does not fix #270 at all : 
 - %z does not exist with msys2
 - %l means long in, but on Windows, long is always a 32 bits integer, so warning on Windows 64 bits

PR #272, opened 2020-06-22
Make CK_ATTRIBUTE_FORMAT refer to the right arguments.

jamesjer:
With the release of version 0.15.0 and the addition of CK_ATTRIBUTE_FORMAT, gcc now complains that every use of a macro with an assert message passes too many arguments for the format.  Part of the problem is some unnecessary passing of trailing NULL arguments.  The main problem, though, is that the CK_ATTRIBUTE_FORMAT directive on `_ck_assert_failed` refers to the wrong arguments.  Argument 3 is `expr`, which is a constant string containing zero format directives.  The argument you want is argument 4, but it cannot be referred to since it is implicit in the triple dots.  The fix is to make it an explicit argument, as in this commit.

### DIFF
--- a/src/check.c
+++ b/src/check.c
@@ -362,17 +362,13 @@ void _mark_point(const char *file, int line)
     send_loc_info(file, line);
 }
 
-void _ck_assert_failed(const char *file, int line, const char *expr, ...)
+void _ck_assert_failed(const char *file, int line, const char *expr,
+                       const char *msg, ...)
 {
-    const char *msg;
-    va_list ap;
     char buf[BUFSIZ];
     const char *to_send;
 
     send_loc_info(file, line);
-
-    va_start(ap, expr);
-    msg = (const char *)va_arg(ap, char *);
 
     /*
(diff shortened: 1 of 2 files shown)


brarcher:
Thanks for the fix!

marcelhollerbach:
This commit breaks our usage of check in efl. The log of it:
https://travis-ci.org/github/Enlightenment/efl/jobs/711611763#L2401

I researched a bit in a few opensource projects, and it seems that the usage of "fail_if(exp)" without a message behind it is used quite often. (efl, mesa, pulseaudio) Which do not work with the 0.15.1 release...

brarcher:
Thanks for pointing this out. A fix for this has been [merged](https://github.com/libcheck/check/pull/298) into master now. I plan on cutting a release with just that fix in the coming days. In the meantime, if you would not mind take a look at the fix and see if it would work for you. If it would not, let me know.

marcelhollerbach:
Build break fixed here as well, thank you very much.

PR #273, opened 2020-06-22
Improve the info directory entry

jamesjer:
The main info directory on many Linux distributions contains one line per package, with the identify of the package on the left, and a 1-line summary of its purpose starting at (zero-based) column 32.  This change makes the info entry fit in better on such distributions.


brarcher:
Ah, I see what you mean. The [texinfo docs](https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Installing-Dir-Entries.html) mention this. Thanks for the fix.

Issue #293, opened 2020-07-25
fail_if broken with check 0.15.1


buscher:
fail_if is broken with check-0.15.1 (did not test 0.15.0)

The following code no longer compiles
```
#include <check.h>

void test_check0151()
{
    int a = 1, b = 2;
    fail_if(a == b);
}
```

with the output
```
test_check0151.c: In function 'test_check0151':
test_check0151.c:8:5: error: too few arguments to function '_ck_assert_failed'
    8 |     fail_if(a == b);
      |     ^~~~~~~
/usr/include/check.h:502:27: note: declared here
  502 | CK_DLL_EXP void CK_EXPORT _ck_assert_failed(const char *file, int line,
      |                           ^~~~~~~~~~~~~~~~~
```

I know fail_if is deprecated like forever... but old code is still using it :/

brarcher:
Ah, I see. This was probably broken in 0.15.0 as well. Check has a test for fail_if and passes a NULL as a second argument. There is a [comment around there](https://github.com/libcheck/check/blob/c94962c5ed3120291557e978fa523780a802ac93/tests/check_check_sub.c#L121) as to why::
```
/* taking out the NULL provokes an ISO C99 warning in GCC */
```

The changes in 0.15.0 enabled the checking of the printf style arguments, and in doing so likely now points out when no arguments are provided, such as with `fail_if(a == b)`.

The [documentation](https://github.com/libcheck/check/blob/184ea1798a6ecc67c20378b557ddb66036ee8fbb/doc/check.texi#L1265) on the deprecated fail_if mention:

```
(Deprecated) Fails test if supplied condition evaluates to true and
displays user provided message.
```

The intent appears to be to have callers pass in a message, rather than making it optional. Up to 0.15.0 it was not enforced, and now a missing message is being pointed out.

I tried adding something like this:
```
#define fail_if(expr) fail_if(expr, NULL)
```
to see if it could be worked around, but this re-defines the macro rather than provides an alternative, so it will not work.

There are a few ways to solve this:
1. Uses of fail_if() that have no message are given a NULL message (requires client code to change)
2. Printf style warnings introduced in 0.15.0 are removed

I'd rather not remove the new warnings as they will help code use the correct printf arguments for Check's APIs. Do you have some other suggestions that may work?

mandree:
[...]

PR #298, opened 2020-08-02
Add extra NULL argument at the end of fail* APIs

brarcher:
The fail, fail_unless, and fail_if APIs were expected to contain
a message explaining the failure. This was never enforced, and
it was possible to write unit tests without providing messages.

In #249 a change was introduced to add printf argument checking to the 
Check assert APIs, including the fail APIs. There were a few fixes for this
in http://github.com/libcheck/check/releases/tag/0.15.1. Those changes
proved problematic for the uses of the fail* APIs without arguments,
as those uses were now flagged as missing the necessary arguments.

A fix proposed by heftig in http://github.com/libcheck/check/issues/293
is to add a new NULL to the end of every fail* call in the macro
itself. For users of these APIs who do pass a message there will
be a new warning about too many arguments. As the fail APIs are
deprecated, this new warning is a reasonable trade-off, and can
be avoided by switching fail* calls to ck_assert* calls.

If a fail* APIs is passing a message, the emitted warning might be:
```
check_check_sub.c:65:23: warning: too many arguments for format [-Wformat-extra-args]
   fail_unless(1 == 2, "This test should fail");
                       ^
../src/check.h:472:77: note: in definition of macro ‘fail_unless’
     _ck_assert_failed(__FILE__, __LINE__, "Assertion '"#expr"' failed" , ## __VA_ARGS__, NULL)
                                                                             ^~~~~~~~~~~
```

### DIFF
--- a/src/check.h.in
+++ b/src/check.h.in
@@ -466,7 +466,10 @@ static void __testname ## _fn (int _i CK_ATTRIBUTE_UNUSED)
  *
  * This call is deprecated.
  */
-#define fail_unless ck_assert_msg
+#define fail_unless(expr, ...) \
+  (expr) ? \
+    _mark_point(__FILE__, __LINE__) : \
+    _ck_assert_failed(__FILE__, __LINE__, "Assertion '"#expr"' failed" , ## __VA_ARGS__, NULL)
 
 /*
  * Fail the test case if expr is false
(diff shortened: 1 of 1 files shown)


mikkoi:
The commit is question is "Add macro CK_ATTRIBUTE_FORMAT"  https://github.com/libcheck/check/pull/249/commits/5e2b32fccc8c4c2440fdf8eccd0eb73fc5a3dcd9

I wanted to use GCC's facilities (since they are available).
But this fix is for all users, not just GCC users. It might react differently in different compilers.

But if that part of the API is deprecated anyway, it might work even if it might cause some extra headache.

Otherwise, just get rid of my commit "Add macro CK_ATTRIBUTE_FORMAT". It was wishful thinking on my part. :-)

PR #305, opened 2020-10-11
Windows: fix size prefix specifiers for printf()

vtorri:
On Windows (mingw(-w64) and old Visual Studio, %z and %t are not
supported. Use the size prefix specifiers for printf  in msvcrt.dll. See :
https://docs.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=vs-2019

### DIFF
--- a/lib/libcompat.h
+++ b/lib/libcompat.h
@@ -89,6 +89,16 @@
 #include <process.h>            /* getpid */
 #endif /* _MSC_VER */
 
+#ifdef _WIN32
+#define CK_FMT_ZU "%Iu"
+#define CK_FMT_ZD "%Id"
+#define CK_FMT_TD "%Id"
+#else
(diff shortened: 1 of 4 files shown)


lib/libcompat.h
```diff
@@ -89,6 +89,16 @@
 #include <process.h>            /* getpid */
 #endif /* _MSC_VER */
 
+#ifdef _WIN32
+#define CK_FMT_ZU "%Iu"
```
brarcher, line 93: Maybe leave a comment here saying why this is necessary, and maybe give a URL
as a reference.

brarcher:
Thanks for the change!

Would you mind putting some of the commit message comment next to the `#define` for documentation?

vtorri:
done

Issue #10, opened 2015-12-25
Website: 'NEWS' tab causes download


njh:
When I click on the 'NEWS' tab on the check website, in Chrome on Mac OS X, it causes it to download the file instead of viewing it.

Fetching the page with curl shows that the MIME type is 'application/octet-stream'.

    $ curl -v 'http://libcheck.github.io/check/NEWS'
    *   Trying 185.31.19.133...
    * Connected to libcheck.github.io (185.31.19.133) port 80 (#0)
    > GET /check/NEWS HTTP/1.1
    > Host: libcheck.github.io
    > User-Agent: curl/7.44.0
    > Accept: */*
    > 
    < HTTP/1.1 200 OK
    < Server: GitHub.com
    < Content-Type: application/octet-stream
    < Last-Modified: Thu, 24 Dec 2015 22:16:38 GMT
    < Access-Control-Allow-Origin: *
    < Expires: Fri, 25 Dec 2015 23:53:16 GMT
    < Cache-Control: max-age=600
    < X-GitHub-Request-Id: B91F1319:2F47:22CE40CE:567DD493
    < Content-Length: 20291
    < Accept-Ranges: bytes
    < Date: Fri, 25 Dec 2015 23:43:16 GMT
    < Via: 1.1 varnish
    < Age: 0
    < Connection: keep-alive
    < X-Served-By: cache-lhr6333-LHR
    < X-Cache: MISS
    < X-Cache-Hits: 0
    < X-Timer: S1451086996.058975,VS0,VE197
    < Vary: Accept-Encoding
    < X-Fastly-Request-ID: b7683b36740a99519ba7343452e15b5ffceae2c4
    <

brarcher:
Well, that stinks. The same thing happens with the LGPL license, and also on other browsers besides Chrome.

According to this:

http://stackoverflow.com/questions/15951012/can-mime-types-of-github-pages-files-be-configured

it seems that GitHub is unable to determine the mimetype of the files because they do not have an extension, and fall back to application/octet-stream. 

I attempted to set the target of the link to open in a new browser tab, hoping that this would convince a browser to not download the file. 

http://www.w3schools.com/tags/att_a_target.asp

However, the browser still downloaded the file. Example here:

http://brarcher.github.io/check/

Without adding the extension of these files, I'm not sure of a way to resolve this, as the issue is on GitHub's side.

ladar:
[...]

PR #308, opened 2020-10-24
Issue 10 website news tab causes download

jflopezfernandez:
## Commit
As the related issue documents, GitHub pages cannot
determine the file type of the NEWS or LGPL files because
they do not have a file extension, so the server uses a
default MIME type of application/octet-stream, which
triggers a download, rather than letting the user actually
read the file.

To circumvent this problem, I changed to file links from
relative links to full hyperlinks referencing the GitHub
repository URL instead. The NEWS file is now viewable via
the "raw" view option accessible through GitHub, and the
LGPL file links to the GitHub license file default.

I chose not to simply use the raw view for the license file
since I felt most people looking at that file would
probably not be lawyers, and would therefore benefit from
the summary GitHub displays at the top of the page.

## Notes
I tested the links using the Firefox web developer tools.

---

This pull request fixes issue #10


jflopezfernandez:
This pull request sought to merge the changes to the master branch instead of the GitHub pages branch.

PR #309, opened 2020-10-24
Fix GitHub Pages Site Download Bug #10

jflopezfernandez:
## Commit
As the related issue documents, GitHub pages cannot
determine the file type of the NEWS or LGPL files because
they do not have a file extension, so the server uses a
default MIME type of application/octet-stream, which
triggers a download, rather than letting the user actually
read the file.

To circumvent this problem, I changed to file links from
relative links to full hyperlinks referencing the GitHub
repository URL instead. The NEWS file is now viewable via
the "raw" view option accessible through GitHub, and the
LGPL file links to the GitHub license file default.

I chose not to simply use the raw view for the license file
since I felt most people looking at that file would
probably not be lawyers, and would therefore benefit from
the summary GitHub displays at the top of the page.

## Notes
I tested the links using the Firefox web developer tools.

---

This pull request fixes issue #10
Pull request #308 incorrectly attempted to merge change to master and has been deleted

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,7 +39,7 @@ http://creativecommons.org/licenses/by-nc/2.5/
 
 <ul id="navlist">
 <li id="active"><a href="#" id="current">Home</a></li>
-<li><a href="./NEWS">NEWS</a></li>
+<li><a href="https://raw.githubusercontent.com/libcheck/check/master/NEWS">NEWS</a></li>
 <li><a href="./web/install.html">Install</a></li>
 <li><a href="./doc/doxygen/html/check_8h.html">API</a></li>
 <li><a href="./doc/check_html/check_3.html">Tutorial</a></li>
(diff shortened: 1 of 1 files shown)


brarcher:
It may be better to fix this in the master branch first. During a release, the master branch is merged into the gh-pages branch, and it will pick up the change then.

https://github.com/libcheck/check/blob/master/HACKING#L118

Or, just put the same change in the master branch too, so it is correct in both places. Whichever is fine.

brarcher:
Oh, and don't worry about the AppVeyor check failure. Since GitHub released its Actions system and added Windows VMs the [Windows testing](https://github.com/libcheck/check/blob/master/.github/workflows/windows.yml) has been moved from AppVeyor to GitHub Actions. I've disabled AppVeyor from running now, as it should no longer be necessary.

jflopezfernandez:
> It may be better to fix this in the master branch first. During a release, the master branch is merged into the gh-pages branch, and it will pick up the change then.
> 
> https://github.com/libcheck/check/blob/master/HACKING#L118
> 
> Or, just put the same change in the master branch too, so it is correct in both places. Whichever is fine.

Will do. I did initially by accident, but the I assumed from the amount of changes in the merge we were using a completely separate branch for GitHub pages. I'll go ahead and make the change.

> Oh, and don't worry about the AppVeyor check failure. Since GitHub released its Actions system and added Windows VMs the [Windows testing](https://github.com/libcheck/check/blob/master/.github/workflows/windows.yml) has been moved from AppVeyor to GitHub Actions. I've disabled AppVeyor from running now, as it should no longer be necessary.

Yea, I was wondering about that, thank you!

PR #315, opened 2020-11-02
Docs/create community files

jflopezfernandez:
I created an `ISSUE_TEMPLATE` directory in the `.github` project directory to allow us to pre-populate bug-report related issues and speed up the turnaround time by being able to skip the information request phase, letting us get right into the verification, diagnosis, fix, and test phases.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Check Library Users Mailing List
+    url: check-users@sourceforge.net
+    about: User questions are accepted and welcome!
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Check Library Users Mailing List
+    url: check-users@sourceforge.net
```
brarcher, line 4: Should we direct users to post questions on GitHub rather that the mailing list?

Sometimes users include attachments with their questions (like screenshots). if they are too big an admin of the mailing list needs to approve it. It would be easier if the questions (and their attachments) were handled by GitHub. It may be easier to discover similar questions if they are all on GitHub as well.

jflopezfernandez: Yea, I agree; I saw the configuration file allowed for including the development mailing list and I thought it might look nice and official, but I think including it might be counterproductive. I didn't know about the mailing list attachment size limit, either.

PR #319, opened 2020-12-27
check_mem_leaks: integrate check_mem_leaks into CMake

hvarga:
Commit 8699473b65011b50766d536222329eeb3953f1cc initialy introduced a
new test, check_mem_leaks, that can be used with Valgrind to find memory
errors.

Since then, test check_mem_leaks can be built with GNU Autotools scripts
but lacks integration into CMake scripts. This means that, when CMake is
used to build the project, check_mem_leaks will NOT be built.

This commit resolves this by adding support for building check_mem_leaks
into CMake scripts as well.

Signed-off-by: Hrvoje Varga <hrvoje.varga@gmail.com>


brarcher:
Thanks for adding the test program to the CMake build!

PR #320, opened 2020-12-27
Resolve all build warnings with too many arguments to _ck_assert_failed

hvarga:


brarcher:
Adding the extra NULLs at the end was a trade-off to keep compatibility with the deprecated fail*() APIs after a printf-style check was added to all the APIs. See https://github.com/libcheck/check/pull/298 for details.

hvarga:
Ah, I see. Thanks for the clarification. Based on this input, I am closing this one down since it is not needed.

PR #322, opened 2021-01-18
PowerPC test suite failures

cgzones:
On PowerPC architectures (ppc, ppc64el, powerp) 'long double' has a
precision of 106-bit, compared to 80-bit precision on amd64.

This leads to the test_ck_assert_(float|double|ldouble)_eq_tol succeed
rather than fail as expected, cause 0.003-0.002 will be actually
slightly bigger than 0.001 and not slightly smaller.

Increase the change to the tolerance, so it will be on all architectures
smaller than the difference of ~0.001 and the unit tests will fail as
expected.


brarcher:
I appreciate your identifying this issue and sending a patch our way. 

I'm hesitant, as none of the platforms used for automated testing are PowerPC, so it is possible that this could be silently broken again in the future. The testing issue will be resolved for today, at least.

PR #339, opened 2021-10-18
Windows: Inform compiler (GCC/MinGW) that format strings to use are GNU/c99 style

bradh352:
Check uses a private snprintf implementation that does not support the Windows format strings of %I64u or similar and instead support the C99 and GNU format format strings like %llu and %zu.  Let the compiler know this otherwise we'll get warnings every time %llu, %lld, %zu are used.


brarcher:
Thanks!

PR #364, opened 2025-10-25
Fix compilation using MinGW-W64 13 and later

dgburr:
Version 13 of MinGW-W64 changed the declaration of `clock_gettime()` in `pthread_time.h` to a function definition which calls either `clock_gettime32()` or `clock_gettime64()`, see https://github.com/mingw-w64/mingw-w64/blob/v13.x/mingw-w64-libraries/winpthreads/include/pthread_time.h. This conflicts with check's definition of `clock_gettime()` in `clock_gettime.c`.  In order to prevent this, we can define `WIN_PTHREADS_TIME_H` in `libcompat.h` to prevent the contents of `pthread_time.h` from being included.

### DIFF
--- a/lib/libcompat.h
+++ b/lib/libcompat.h
@@ -143,6 +143,7 @@ extern int fpclassify(double d);
 
 /* provides localtime and struct tm */
 #ifdef HAVE_SYS_TIME_H
+#define WIN_PTHREADS_TIME_H 1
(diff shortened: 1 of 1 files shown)


wissem01chiha:
Hi, can someone review this PR ? some issues in other projects fail to compile with mingw  @eschaton ? 
thanks

PR #365, opened 2026-06-14
Fix loop variable name

buscher:
- START_TEST uses "int _i", not "i"


eschaton:
Looks good to me.

brarcher:
Thanks!

PR #367, opened 2026-06-25
Fix tr_rtype return type

eschaton:
Right now the return type of the `tr_rtype()` API is `int`. Since there's an `enum test_result` that `rtype` is defined in terms of elsewhere, the API should also use that.


brarcher:
Thanks!

PR #368, opened 2026-06-25
Add tr_tname API

eschaton:
Right now you can only get the test case name from a `TestResult`, using the `tr_tcname()` API. This means if you're trying to hook Check's result reporting into some other system, you can only get half the information you need to describe a test's success or failure. Since `TestResult` tracks the test name too, this can be easily remedied by also exposing a `tr_tname()` API.


brarcher:
Thanks!

PR #376, opened 2026-07-01
Get macOS workflows working again

eschaton:
I'm using this PR to try to get macOS workflows working. I expect this will require a combination of adjustment to the workflows themselves as well as to the `configure.ac` file to modernize both.


eschaton:
Looks like the only remaining issue is that processing `check.texi` is generating errors that aren't generated on Linux.

brarcher:
I see the same check.texi issues in the windows builds too. I'm thinking the version of texinfo tool used in the Linux tests is more forgiving of errors. 

I've updated check.text to pass, and fixed the other issues with the windows tests. That's now merged into master. Feel free to rebase to pick up those changes and confirm that all the osx issues are resolved.

eschaton:
Excellent, thanks! Hopefully just a little more poking at it and it'll all work!

eschaton:
OK, now we're down to only `build_osx_cmake` failing. That should be tractable.

eschaton:
@brarcher Where is `HAVE_SETITIMER` supposed to come from in a CMake build? It's not being set and the `#else` in both `timer_{delete,settime}` is missing a `return 0;` which can cause the call to return an error, not 0.

brarcher:
> Where is HAVE_SETITIMER supposed to come from in a CMake build? It's not being set and the #else in both timer_{delete,settime} is missing a return 0; which can cause the call to return an error, not 0.

Thanks for catching that (I missed the return when fixing the windows builds earlier). 
I see you also figured out where the HAVE_SETITIMER should go for cmake.

Thanks for resolving the osx builds! Now the builds are back to green

PR #378, opened 2026-07-03
WIP: Introspection APIs (#366)

### LINKED ISSUE #366: Introspection APIs
I've been doing ANSI C projects for fun the past couple of years, and I'll often start out writing my tests using XCTest in Xcode out of convenience, and then migrating them to Check for wider platform support. (I worked on Xcode for a couple decades before retiring a couple of years ago, and was responsible for integrating testing support into Xcode way back in Xcode 2.1…)

Like other IDEs, Xcode's test infrastructure is at least somewhat amenable to a degree of integration with Check. However, to get better integration than just catching failures, Check will need some additional APIs beyond getting  a TestResult collection after a run. In particular, I would need to be able to iterate over the `Suite` and `Test` instances in a `Suite` _before_ running them, and to be able to run an individual `Test` or `Suite`.

Having these APIs would allow me to create XCTestSuite and XCTestCase subclasses that can (at setup time) generate further synthetic subclasses and methods at runtime that can be used to map Xcode's runtime test specifiers to the specific Check `Suite` and `Test` to invoke, and then report their state back to Xcode in a way it'll present as success or failure with proper source annotation. (None of this should require an Xcode plug-in or anything of the sort, it should all be doable within the XCTest API if Check itself has the necessary API.)

eschaton:
Adding introspection APIs for test suite runners, suites, test cases, and tests.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,9 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf-8
+indent_size = 4
+indent_style = space
+tab_width = 8
 
 [makefile,Makefile]
 indent_style = tab
--- a/src/check.c
+++ b/src/check.c
@@ -236,6 +236,35 @@ void tcase_set_tags(TCase * tc, const char *tags_orig)
     tc->tags = tag_string_to_list(tags_orig);
 }
 
+int tcase_ntags(TCase * tc)
+{
+    if(tc == NULL)
+        return 0;
+
+    return check_list_length(tc->tags);
+}
+
+const char **tcase_tags(TCase * tc)
+{
+    int i = 0;
+    List *l;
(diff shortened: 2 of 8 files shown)


eschaton:
Anyone have any idea what the test failures here might be, based on eyeballing my commits? I expect it's related to my change to make `test_name()` take a `TCase` (so `tcase_current()` is needed to get the current one within a test).

eschaton:
Turns out one just can't rely on being able to get a filled-in `TCase` during test execution, so instead of using `tcase_name(tcase_current())` I've just added `tcase_current_name()` so `tcase_name()` can still take a `TCase *`.

eschaton:
Remaining to-do items:
- A few documentation comments for new API
- Tests for new API

